PR #3, opened 2022-04-28
Configure wallet i18n

soerenbf:
## Purpose

Setup internationalization (i18n) for browser wallet project. The framework chosen to handle this is documented here: https://www.i18next.com and https://react.i18next.com

The initial thoughts with regards to how translation files are structured are:
- 1 translation namespace for everything 'shared', which is located in the `shared` folder (used for general purpose components)
- 1 translation namespace per page (`Account`, `SendTransaction`, etc...)
- 1 translation namespace per page-layout (`MainLayout`, `FullscreenPromptLayout`, etc...)
- Everything is bundled in a locale file for each language supported in `shell/i18n/locales/<lang>.ts`

Dependent on #2 being merged.

## Changes
- Added i18n framework
- Added translations file structure
- Added translations for text added so far
- Added type extension for translation framework, to ensure strong typing of texts

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/types/react-i18next.d.ts
```diff
@@ -0,0 +1,15 @@
+// import the original type declarations
+import 'react-i18next';
+
+import type { resources, defaultNS } from '../popup/shell/i18n';
+
+// react-i18next versions higher than 11.11.0
+declare module 'react-i18next' {
```
shjortConcordium, line 7: Why do you need to add this?
Does the package not have proper ts support? :thinking:

soerenbf: They do, this is a way to extend their types to include the specific translations for the project. It's how they documented how to do this on their website. Without this, selecting a translation using any method will not be strongly typed, and it is when this is included.

shjortConcordium (approved):
LGTM, Are you thinking we are also implementing danish along the way (useful for testing, I guess)?

soerenbf:
> LGTM, Are you thinking we are also implementing danish along the well (useful for testing, I guess)?

Yes, my thought was, that it's useful for testing. Should also be easy enough for us to just c/p the changes to the english locale and quickly write up some stuff for danish as well. We can always remove it in the root if we don't want it.

packages/browser-wallet/src/popup/shared/i18n/en.ts
```diff
@@ -0,0 +1,5 @@
+const t = {
+    test: 'Testing....',
```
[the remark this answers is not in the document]
soerenbf: Using it in a future PR.. I had it in there to exemplify structure for text for shared components. Will leave it here for now :smiley:

PR #4, opened 2022-04-29
Configure wallet forms

soerenbf:
## Purpose

Adds a forms module to be used for handling and validating user input.

Builds on top of the changes implemented in #3.

## Changes

- Add forms library `react-hook-form`
- Add `Form` and `Input` components
- Add HoC's for making both controlled and uncontrolled inputs compatible with the `Form` component.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/shared/utils/basicHelpers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const notNull = <T>(v: T | null | undefined): v is T => v != null;
+
+export function noOp(): void {
+    return undefined;
+}
+
+export async function asyncNoOp(): Promise<void> {
+    // eslint-disable-next-line no-promise-executor-return
+    return new Promise((resolve) => resolve());
```
shjortConcordium, line 13:
```suggestion
    return Promise.resolve();
```

packages/browser-wallet/src/shared/utils/basicHelpers.ts
```diff
@@ -0,0 +1,14 @@
+export function isDefined<T>(v?: T): v is T {
+    return v !== undefined;
+}
+
+export const notNull = <T>(v: T | null | undefined): v is T => v != null;
+
+export function noOp(): void {
+    return undefined;
+}
```
shjortConcordium, line 9: :thinking: 
We might need some type-checking on this project?

shjortConcordium: As far as I understand, undefined != void

packages/browser-wallet/src/popup/shared/Form/common/utils.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { RequiredControlledFieldProps, RequiredUncontrolledFieldProps } from './types';
+
+type MakeControlledProps<TFieldValues extends FieldValues = FieldValues> = Omit<
+    MakeRequired<ControllerProps<TFieldValues>, 'control'>,
+    'render'
+>;
+
+/**
+ * @description
+ * Can be used to create form fields for controlled fields, i.e. where field state is kept outside..
```
shjortConcordium, line 24:
```suggestion
 * Can be used to create form fields for controlled fields, i.e. where field state is kept outside.
```

packages/browser-wallet/src/popup/shared/Form/Form.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.form-input {
+    @include when-invalid {
+        .form-input__field {
+            border-color: red;
+        }
+    }
+}
+
+.form-error-message {
```
shjortConcordium, line 15: So in the desktop wallet we use camel case, is this some special case, or are you suggesting we change to kebab case in this project?

shjortConcordium (commented):
How much of this Form stuff is copied from the DW?
(Should we at some point consider making a shared repo for this stuff?)

soerenbf:
> How much of this Form stuff is copied from the DW? (Should we at some point consider making a shared repo for this stuff?)

Would make sense to create a shared repo. This project relies on a newer version than the DW, and required some changes for it to work, so it would require updating the DW form fields to adapt to this.

packages/browser-wallet/src/shared/utils/basicHelpers.ts
soerenbf: what do you mean we might need type checking? I agree that it doesn't make sense, but apparently typescript accepts it :shrug: 

I'll correct it to an empty block instead.

packages/browser-wallet/src/popup/shared/Form/Form.scss
soerenbf: It essentially attempts to do the same as css modules, i.e. scope css rules to modules (or blocks in this case). Without css modules, it's easier to makes css rules that apply to blocks inside other blocks (override styling of nested components in a special case), as you don't have the automatic hashed scoping from css modules.

packages/browser-wallet/src/shared/utils/basicHelpers.ts
```diff
@@ -0,0 +1,12 @@
+export function isDefined<T>(v?: T): v is T {
+    return v !== undefined;
+}
+
+export const notNull = <T>(v: T | null | undefined): v is T => v != null;
+
+export function noOp(): void {}
+
+export async function asyncNoOp(): Promise<void> {
+    // eslint-disable-next-line no-promise-executor-return
```
shjortConcordium, line 10:
```suggestion
```
Didn't even notice it before

shjortConcordium (approved):

shjortConcordium:
> > How much of this Form stuff is copied from the DW? (Should we at some point consider making a shared repo for this stuff?)
> 
> Would make sense to create a shared repo. This project relies on a newer version than the DW, and required some changes for it to work, so it would require updating the DW form fields to adapt to this.

I guess we should keep it in mind, but obviously not for pre-mvp.

PR #7, opened 2022-05-17
Allow connection from dApp

soerenbf:
## Purpose

Adds functionality to check, handle, and store connection attempts from wallet API consumers.

## Changes

- Refactored messaging to simpler promise based request/response fashion.
- Added `example.html` in workspace root to test wallet API
- Moved chrome storage functionality to shared folder for reuse between background/popup
- Added functionality to detect existing windows and only spawn new windows when needed
- Added message based routing

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/pages/ConnectionRequest/i18n/da.ts
```diff
@@ -0,0 +1,12 @@
+import type en from './en';
+
+const t: typeof en = {
+    title: 'Forbindelsesforespørgsel',
```
shjortConcordium, line 4: :tophat: 
:monocle_face:

packages/browser-wallet/src/popup/page-layouts/FullscreenPromptLayout/FullscreenPromptLayout.tsx
```diff
@@ -1,17 +1,77 @@
...
-        <>
-            <header>
-                <Link to={absoluteRoutes.home.path}>X</Link>
-            </header>
-            <main>
-                <Outlet />
-            </main>
-        </>
+        // eslint-disable-next-line react/jsx-no-constructed-context-values
+        <fullscreenPromptContext.Provider value={{ onClose, withClose }}>
```
shjortConcordium, line 66: Is this not the exact use case, why the rule exists?

shjortConcordium: or I guess withClose is recreated each render anyway?

packages/browser-wallet/src/inject/index.ts
```diff
@@ -1,8 +1,10 @@
...
-import { logger } from '@concordium/browser-wallet-message-hub/src/message-handlers/logger';
 
 // Inject WalletAPI to dApp
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 (window as any).concordium = walletApi as IWalletApi;
 
-logger.log('::InjectedScripted executed');
+// TODO figure out best way to let DOM know that the API is ready for use. This assumes consumers declare a global function "concordiumReady"
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(window as any).concordiumReady();
```
shjortConcordium, line 10: Is when the wallet API is ready? (I thought it would just be when window.concordium wasn't undefined? (but I guess this is to have a callback?)

packages/browser-wallet/src/inject/index.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable no-console */
```
shjortConcordium, line 1:
```suggestion
```

packages/browser-wallet/src/background/window-management.ts
```diff
@@ -0,0 +1,71 @@
+import { ExtensionMessageHandler, InternalMessageType } from '@concordium/browser-wallet-message-hub';
+
+import { height, width } from '@popup/constants/dimensions';
+import { spawnedPopupUrl } from '@shared/constants/url';
+// eslint-disable-next-line import/no-cycle
+import bgMessageHandler, { onMessage } from './message-handler';
```
shjortConcordium, line 6: Hmm, could we restructure/add a third file, so we can avoid this?

packages/browser-wallet-api/src/wallet-api.ts
```diff
@@ -1,79 +1,85 @@
...
+type WalletEventHandler<T = any> = (payload: T) => void;
 
 export interface IWalletApi {
-    sendTransaction(): Promise<string>;
-    signMessage(): Promise<Message>;
-    getAccounts(): Promise<Message>;
+    addChangeAccountListener(handler: WalletEventHandler<string>): void;
+    sendTransaction(): Promise<boolean>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    signMessage(): Promise<any>;
```
shjortConcordium, line 15: These signatures will be changed, but I guess we'll do it, when we actually implement the features :upside_down_face:

shjortConcordium (commented):
Have some initial comments, will play with it tomorrow.

packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -1,26 +1,42 @@
...
+    [ChromeStorageKey.JsonRpcUrl]: storedJsonRpcUrl,
+    [ChromeStorageKey.UrlWhitelist]: storedUrlWhitelist,
+};
 
 /**
  * @description
  * Create an atom that automatically syncs with chrome local storage.
  */
 export const atomWithChromeStorage = <V>(key: ChromeStorageKey, fallback: V) => {
+    const accessor = accessorMap[key] as unknown as StorageAccessor<V>;
```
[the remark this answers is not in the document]
soerenbf: Well, if it's not there you get a union of the different "accessors", which is obviously not accepted by the compiler. I don't know how to avoid the cast. Do you have any ideas? 😄

packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -1,26 +1,42 @@
...
+
+    if (accessor === undefined) {
+        throw new Error(`Could not find storage for key: ${key}`);
+    }
+
+    const { get: getStoredValue, set: setStoredValue } = accessor;
```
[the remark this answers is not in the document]
soerenbf: The chrome storage is accessible from all contexts of the extensions, except the injected script, which is part of the web page it is injected onto. The accessor stuff is to ensure consistency in how stored values are accessed across the different contexts.

packages/browser-wallet/src/popup/page-layouts/FullscreenPromptLayout/FullscreenPromptLayout.tsx
soerenbf: I'll give it another shot. I guess the close is recreated every render, and that cascades to the other stuff, so maybe I forgot something.

packages/browser-wallet/src/inject/index.ts
soerenbf: Yes, otherwise the consumer would have to periodically check - I think this is more elegant, but open to other suggestions. You can check `example.html` how it's used.

packages/browser-wallet-api/src/wallet-api.ts
soerenbf: Yes, that was my thought as well.

packages/browser-wallet/src/background/window-management.ts
soerenbf: I'll see what I can do! ⛑️

packages/browser-wallet/src/inject/index.ts
shjortConcordium: The only issue I have is that if the wallet is ready before the consumer is ready, would this not give an error?
(and then when they define it afterwards, it wouldn't get triggered?)

shjortConcordium:
If I have sendTransaction open and I change tab, I get:
```
background.js:1 Uncaught (in promise) Error: The message port closed before a response was received.
```
We probably don't need to address that now, but we should be aware?

soerenbf:
> If I have sendTransaction open and I change tab, I get (in the extension console):
> 
> ```
> background.js:1 Uncaught (in promise) Error: The message port closed before a response was received.
> ```
> 
> We probably don't need to address that now, but we should be aware?

Added some more error handling to the different contexts in the latest commit.

shjortConcordium (approved):

PR #10, opened 2022-06-01
Allow signing messages

shjortConcordium:
## Purpose

Allow signing a message.

## Changes

Implements the flow for signing a message, and returning the signature.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api/src/wallet-api.ts
+++ b/packages/browser-wallet-api/src/wallet-api.ts
@@ -35,8 +35,12 @@ export interface IWalletApi {
      * @param payload the payload of the transaction to be signed and sent.
      */
     sendTransaction(type: AccountTransactionType, payload: AccountTransactionPayload): Promise<string>;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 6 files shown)


soerenbf (approved):
We should include a better example though, but I think it's better to add this to the backlog instead of squeezing it in.

packages/browser-wallet-api/src/wallet-api.ts
```diff
@@ -57,9 +59,16 @@ class WalletApi implements IWalletApi {
     /**
      * Sends a sign request to the Concordium Wallet and awaits the users action
      */
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    public signMessage(): Promise<any> {
-        return this.sendMessage(MessageType.SignMessage, {});
+    public async signMessage(message: string): Promise<AccountTransactionSignature> {
```
soerenbf, line 62: `IWalletAPI` interface needs to be changed to match this.

shjortConcordium: :weary: good catch.

PR #12, opened 2022-06-02
Wallet api types package

soerenbf:
## Purpose

Provide types for apps interacting with the Concordium wallet, without having to include code that is already injected into the app by the wallet.

The package added is meant to be published.

## Changes

- Add package with types for the wallet API
- Use types from added package in the other wallet packages where applicable.
- Cleaned up references, so only packages meant to be published are workspace dependencies, i.e. linked into the root node modules folder.

## Checklist

      hard-to-understand areas.


package.json
```diff
@@ -30,8 +30,7 @@
         ]
     },
     "dependencies": {
-        "@concordium/browser-wallet-api": "workspace:*",
-        "@concordium/browser-wallet-message-hub": "workspace:*"
+        "@concordium/browser-wallet-api-types": "workspace:*"
```
shjortConcordium, line 33: Does this actually do anything?

packages/browser-wallet-api-types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+        schema: string
+    ): Promise<string>;
+    /**
+     * Sends a transaction to the Concordium Wallet and awaits the users action. Note that a header is not sent, and will be constructed by the wallet itself.
+     * Note that if the user rejects signing the transaction, this will throw an error.
+     * @param type the type of transaction that is to be signed and sent.
+     * @param payload the payload of the transaction to be signed and sent.
+     */
+    sendTransaction(type: AccountTransactionType, payload: AccountTransactionPayload): Promise<string>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
shjortConcordium, line 35:
```suggestion
    /**
     * Sends a message to the Concordium Wallet and awaits the users action. If the user signs the message, this will resolve to the signature.
     * Note that if the user rejects signing the message, this will throw an error.
     * @param message message to be signed. Note that the wallet will prepend some bytes to ensure the message cannot be a transaction
     */
```

shjortConcordium: Guess something was lost in a merge

shjortConcordium (commented):
Should we add the types package as a dependency to the piggybank example?

soerenbf:
> Should we add the types package as a dependency to the piggybank example?

Yes, exactly my thoughts 🙂

package.json
soerenbf: removed it.

packages/browser-wallet/package.json
```diff
@@ -10,8 +10,7 @@
         "watch": "cross-env WATCH=1 yarn build:dev"
     },
     "dependencies": {
-        "@concordium/browser-wallet-api": "0.1.0",
-        "@concordium/browser-wallet-message-hub": "0.1.0",
```
shjortConcordium, line 14: Does the not still refer to types from the message hub? (why are you also removing that?)

shjortConcordium: Because it's not supposed to be published? :thinking:

soerenbf: They're not going to be released as packages. The references remain in the respective tsconfigs.

shjortConcordium (approved):

packages/browser-wallet-api-types/package.json
```diff
@@ -0,0 +1,20 @@
...
+        "type": "git",
+        "url": "https://github.com/Concordium/concordium-browser-wallet"
+    },
+    "author": {
+        "name": "Concordium Software",
+        "email": "support@concordium.software",
+        "url": "https://concordium.com"
+    },
+    "dependencies": {
+        "@concordium/web-sdk": "^0.1.0"
```
shjortConcordium, line 18:
```suggestion
        "@concordium/web-sdk": "^0.1.1"
```

PR #13, opened 2022-06-02
Update README + dependencies + fix example

shjortConcordium:
## Changes

- Changes to wallet-api and wallet-api-types README's
- web-sdk 0.1.0 -> 0.11 everywhere
- Fix broken example


packages/browser-wallet-api-types/README.md
```diff
@@ -1 +1,84 @@
 # browser-wallet-api-types
+
+This package includes the types for the API to be used in web applications for communicating with the Concordium browser wallet.
+
+This package changes the type of the window object, to include the injected api that is available on `window.concordium`.
+
+Note that this package only contains the types and a description of the API, which is rejected by the Concordium browser wallet.
```
soerenbf, line 7: I guess this is what you meant to write?

```suggestion
Note that this package only contains the types and a description of the API, which is injected by the Concordium browser wallet.
```

packages/browser-wallet-api-types/README.md
```diff
@@ -1 +1,84 @@
...
+
+The actual implementation can be found [In the Concordium browser wallet repository.](https://github.com/Concordium/concordium-browser-wallet/tree/main/packages/browser-wallet-api)
+
+## Using the API
```
soerenbf, line 11: I think we should maybe have a section on including the types for the window object? You can have a look at how it's set up in the `tsconfig.json` of the piggybank example.

packages/browser-wallet-api/README.md
```diff
@@ -2,80 +2,7 @@
 
 This package is an implementation of an API to be used in web applications for communicating with the Concordium browser wallet.
 
-## Using the API
```
soerenbf, line 5: Hmm... I don't think I agree on deleting this actually, but we should also avoid duplication. Maybe we should reference the types package from here, or reference the API package from the types package? Leaning most towards having the full document on the package published on NPM.

shjortConcordium: Yes, I agree, that's why I link to the types package below this.

soerenbf (approved):

PR #16, opened 2022-06-07
Visual improvements to piggy bank example

soerenbf:
## Checklist

      hard-to-understand areas.


examples/piggybank/src/Root.tsx
```diff
@@ -20,12 +23,15 @@ import {
  * it will be necessary to instantiate your own piggy bank using an account available in the browser wallet,
  * and change this constant to match the index of the instance.
  */
-const CONTRACT_INDEX = 5114n; // V0 instance
+const CONTRACT_INDEX = 5122n; // V0 instance
```
shjortConcordium, line 26: I still think this is a little awkward.
But this address is also already smashed?

soerenbf: Yes, smashed it during the demo. I'll create a new one.

I think it's easier to understand if it's hardcoded. You might want to just run the example and see the integration, and not care about whether it's your own piggy bank or not. I think the value from the example comes from being able to easily have something running, instead of having to:

1. Deploy piggy bank
2. Instantiate piggy bank
3. Set up piggy bank with the returned index.

Maybe someone with little SC experience (like myself) comes around, and just wants to see the app code without having to bother with the steps listed above.

shjortConcordium (approved):

PR #18, opened 2022-06-08
UI scaling

soerenbf:
## Purpose

Adds scaling functionality, which attempts to scale the wallet window according to the size of the screen.

Doesn't seem to be completely consistent on a multi-monitor setup when the wallet is opened by clicking the icon (the screen registered is always the primary display, even though the chrome window is on a secondary display). Might be a MacOS only thing.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/styles/config/_functions.scss
```diff
@@ -0,0 +1,9 @@
+@use 'sass:math';
+
+@function strip-unit($value) {
+    @return math.div($value, ($value * 0 + 1));
```
shjortConcordium, line 4: what is x * 0? :thinking: 
I'm confused by this

shjortConcordium:
I don't think I have a problem with the multi-monitors, maybe because they have the same size, though.
It does seem that the text have smaller, even though the size of the window is bigger.

packages/browser-wallet/src/popup/styles/config/_functions.scss
soerenbf: It's a commonly known hackish solution to strip any number value from it's units, f.x. `strip-unit(18px) -> 18`.

soerenbf:
> I don't think I have a problem with the multi-monitors, maybe because they have the same size, though. It does seem that the text have become smaller, even though the size of the window is bigger.

Yes, it's a problem when running monitors of different sizes.

shjortConcordium (approved):

PR #22, opened 2022-06-14
ID card

soerenbf:
## Purpose

Adds ID card component to wallet

## Changes

- Added ID card component
- Added inline input component
- Added general purpose hooks (from desktop wallet project)

## Checklist

      hard-to-understand areas.


orhoj:
Handling of out-of-bounds text is not very nice as the header, text below the identity name and the icon in the lower right corner will start to move. Their placement should be static and independent of the identity name input. Or the input should be limited such that it can never happen.

packages/browser-wallet/src/popup/shared/IdCard/IdCard.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+    // eslint-disable-next-line react/destructuring-assignment
+    const [name, setName] = useState(args.name);
+    return <IdCard {...args} name={name} onNameChange={setName} />;
+};
+
+export const Pending = Template.bind({});
+Pending.args = {
+    name: 'Name',
+    status: 'pending',
+    provider: <>DTS</>,
```
orhoj, line 21: There should be an example with an image as the provider as that is what the Figma sketches use.

packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+                    )}
+                </>
+            )}
+        </Form>
+    );
+}
+
+type Props = {
+    name: string;
+    status: 'pending' | 'approved' | 'rejected';
```
orhoj, line 71: Two things here:
1. The status of an identity should be defined centrally as an enum, as we will be using that in multiple places in the application.
2. In the desktop wallet we use `confirmed` over `approved` and we should stay consistent to avoid confusing ourselves with any differences between the two terms. (Event though I like the approved terminology better :smile:).

Context: The identity provider actually uses the values: 'pending', 'error' and 'done', but it seems that we went with a translation of these values to something that fits the wallet side better. I also think that `done` would be a bit weird to have as the status internally.

packages/browser-wallet/src/popup/shared/utils/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+            const interval = setInterval(fn, rate);
+            return () => clearInterval(interval);
+        }
+        return () => {};
+    }, [enable, fn, rate]);
+}
+
+/** Hook for getting the current time, optionally taking a refresh frequency in
+ * milliseconds defaulting to 1 second */
+export function useCurrentTime(refreshRate = 1000) {
```
orhoj, line 37: This function and others in this file appear to be dead code. They should be removed and added if needed at a later point, but adding them now risks them just being dead forever.

packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
soerenbf: 1. I agree, however I thought it made better sense to wait until the identity object would be used, and then adapt the `IdCard` component to use the enum.
2. Alright. I'll rename it to confirmed.

PR #23, opened 2022-06-15
Add CI

shjortConcordium:
## Purpose

Setup Actions for the repo.

## Changes

- Added actions.
- Fixed lint / type errors
- Updated piggybank example to one that exists on the new testnet
## Checklist

      hard-to-understand areas.


.github/workflows/build-lint-test.yml
```diff
@@ -0,0 +1,149 @@
...
+
+            - name: Get cached dependencies
+              uses: actions/cache@v2
+              with:
+                  path: |
+                      **/node_modules
+                  key: ${{ runner.os }}-v3-${{ hashFiles('**/yarn.lock') }}
+
+            - name: Build
+              run: yarn build:dev
```
soerenbf, line 57: All these commands are for the browser wallet as far as I see. What about checking the other packages and examples to cover all the code included in the repository? Or am I missing some `yarn workspace` magic stuff?

soerenbf (commented):
A few comments and questions:

soerenbf:
This might be unrelated, but would it make sense to make type checking now as well as a husky hook? Don't know if pre-commit is the right choice, at least not if we end up with a solution that just checks the whole project every time?

.github/workflows/build-lint-test.yml
```diff
@@ -0,0 +1,149 @@
...
+            - uses: actions/setup-node@v2
+              if: steps.fetch-dependencies.outputs.cache-hit != 'true'
+              with:
+                  node-version: '14.16.0'
+
+            - name: Download dependencies
+              if: steps.fetch-dependencies.outputs.cache-hit != 'true'
+              run: yarn
+
+    build:
```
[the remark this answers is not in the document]
shjortConcordium: I thought build is faster, and if it doesn't build, there is no real reason to do any of the other things, right?

.github/workflows/build-lint-test.yml
shjortConcordium: Yeah, I guess it only builds the wallet itself.
But the message hub is used by the wallet, and is the api not also?
But you are correct, I'll try adding it, so that it builds the example as well.

shjortConcordium: And type check.

shjortConcordium:
> This might be unrelated, but would it make sense to make type checking now as well as a husky hook? Don't know if pre-commit is the right choice, at least not if we end up with a solution that just checks the whole project every time?

Yeah, I thought the same, but had some trouble with it, like I assume you did. But I'll try again.

soerenbf (commented):
For reference:

- We should add type check for the `wallet-api-types` package as well.
- Maybe it would make sense to leverage `yarn workspace foreach` commands to make it easier to add new workspaces later and have them automatically (almost) covered by CI?

PR #25, opened 2022-06-16
Use events to notify when Wallet API is ready

orhoj:
## Purpose
Use events to notify dApps when the Concordium Wallet API provider is ready.

## Changes
- Dispatches an initialized event when the provider is loaded into `window`.
- Added a method for detecting the Concordium Wallet API provider. This will function should reside in the web-sdk and will be moved later.
- Updated the examples to use the new way of listening for the Concordium Wallet API provider.

## Checklist

      hard-to-understand areas.


orhoj:
I will update the documentation if we agree on the above method for waiting for the API to be ready.

examples/piggybank/src/Root.tsx
```diff
@@ -43,14 +44,39 @@ const isPiggybankSmashed = (state: PiggyBankState): state is PiggyBankStateSmash
...
+        }
+
+        if (window.concordium) {
+            handleConcordium();
+        } else {
+            window.addEventListener('concordium#initialized', handleConcordium, { once: true });
+            setTimeout(() => {
+                handleConcordium();
+            }, timeout);
+        }
```
soerenbf, line 77: All the `alreadyHandled` stuff shouldn't be necessary as a promise only resolves once. Can be simplified by something like:

```suggestion
        if (window.concordium) {
            resolve(window.concordium);
        } else {
            const t = setTimeout(reject, timeout);
            window.addEventListener('concordium#initialized', () => {
                resolve(window.concordium as WalletApi);
                clearTimeout(t);
            }, { once: true });
        }
```

Probably needs to be implemented locally to not annoy the linter too much.

examples/piggybank/src/Root.tsx
```diff
@@ -198,23 +224,31 @@ function PiggyBank() {
...
+
+                    // Listen for relevant events from the wallet.
+                    provider.addChangeAccountListener(setAccount);
+                })
+                .catch(() => setIsConnected(false));
+        } else {
+            // Not yet connected to the wallet.
+            setIsConnected(false);
+        }
+    }, [provider]);
```
soerenbf, line 250: All this (including the provider state hook) seems a bit unnecessary? Why not just chain the connect and the subsequent stuff directly on `detectConcordiumProvider`?

examples/two-step-transfer/index.html
```diff
@@ -5,12 +5,42 @@
...
+                    if (window.concordium) {
+                        handleConcordium();
+                    } else {
+                        window.addEventListener('concordium#initialized', handleConcordium, { once: true });
+                        setTimeout(() => {
+                            handleConcordium();
+                        }, timeout);
+                    }
+                });
+            }
```
soerenbf, line 34: I guess if you end up changing the corresponding function in the piggy bank example, you should change this one too.

soerenbf (changes requested):
Seems a little weird to me that the actions (smash, deposit) use `window.concordium`, as this would be provided from the SDK in the future? I think we should eliminate `window.concordium` from the example altogether and make the actions get the API from the promise instead.

examples/piggybank/src/Root.tsx
orhoj: I thought that I had a good reason for doing it, but when looking at it now I can't see it. So yes, I will collapse all this.

examples/piggybank/src/Root.tsx
orhoj: I ended up with a mixture of what I had and your suggestion that makes it a bit nicer.

packages/browser-wallet-api-types/README.md
```diff
@@ -2,22 +2,53 @@
...
+This package changes the type of the window object, to include the injected API that is available on `window.concordium`.
 
 Note that this package only contains the types and a description of the API, which is injected by the Concordium browser wallet.
 
 The actual implementation can be found [In the Concordium browser wallet repository.](https://github.com/Concordium/concordium-browser-wallet/tree/main/packages/browser-wallet-api)
 
 ## Using the API
 
-The API is automatically injected into web applications if the Concordium browser wallet extension is installed in the browser. It is made accessible by `window.concordium`. A callback (`concordiumReady`), which needs to be defined globally, is called by the injected script to notify applications that the API is ready for use. The following exemplifies how this can be done.
+The API is automatically injected into web applications if the Concordium browser wallet extension is installed in the browser. To get access to the API use the following helper function (this will be implemented in the web-sdk):
```
shjortConcordium, line 13:
```suggestion
The API is automatically injected into web applications if the Concordium browser wallet extension is installed in the browser. To get access to the API use the following helper function (this will be added to the web-sdk):
```
:thinking:

packages/browser-wallet-api-types/README.md
```diff
@@ -2,22 +2,53 @@
...
+The following exemplifies how accessing the API can be done.
+
+```typescript
+detectConcordiumProvider()
+    .then((provider) => {
+        // The API is ready for use.
+        const accountAddress = provider.connect();
+    })
+    .catch(() => console.log('Connection to the Concordium browser wallet timed out.'));
 ```
```
shjortConcordium, line 52:
```suggestion
```typescript
detectConcordiumProvider()
    .catch(() => console.log('Connection to the Concordium browser wallet timed out.'));
    .then((provider) => {
        // The API is ready for use.
        const accountAddress = provider.connect();
    })
    .catch(() => console.log('Connection to the Concordium browser wallet was rejected.'));
```
```
Or is that too involved?

orhoj: No, what I had was wrong, but so is your example. I made an update to it.

shjortConcordium (approved):

PR #28, opened 2022-06-22
Support dApps sending initialize contract transactions

shjortConcordium:
## Purpose

Allow dApps to send initContract transactions.

## Changes

- Add support for initContract transactions.
- Copy some CCD utility functions from DW (and use them)
 - Updated two-step-transfer for new testnet

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/shared/utils/ccd.ts
```diff
@@ -0,0 +1,79 @@
+import {
```
orhoj, line 1: Avoid duplicating code from the desktop wallet. If we need the utility functions in both projects, then we should make a shared library instead.

examples/two-step-transfer/index.html
```diff
@@ -99,6 +124,7 @@
 
     <body>
         <button id="connect">Connect</button>
+        <button id="sendInit">Initiate two step transfer contract</button>
         <button id="sendTransfer">Send transfer</button>
```
soerenbf, line 128: Should we maybe remove this from the example (perhaps put it somewhere else)?

Could do a "donation" example, which would just be a button saying "Donate to me!" and a send transaction 😁

soerenbf: Might be out of scope of the PR, but should be a quick fix, if you find it relevant.

shjortConcordium: Hmm, I know it's not two-step-transfer, but I was thinking this would become a board of buttons to trigger every endpoint.

shjortConcordium: And it would be the same for the signMessage I assume?

soerenbf: Ok. Might be confusing if you (as an outsider) come looking for examples. If we want a "board of buttons", I would suggest that we instead make an example for that purpose, that isn't mixed with integration with a concrete smart contract.

Can wait for another time though, but maybe we should make an issue for the backlog?

soerenbf (approved):

PR #30, opened 2022-06-23
Detector package

orhoj:
## Purpose
Change the types packages to a helpers package that also contains the Concordium browser wallet API detector. This will be published a new package and then we can deprecate the types package. 

## Changes
- Renamed package.
- Added method for detecting the Concordium browser wallet API.
- Updated examples.
- Updated documentation.

## Checklist

      hard-to-understand areas.


packages/browser-wallet-api-helpers/README.md
```diff
@@ -56,16 +32,6 @@ detectConcordiumProvider()
...
-To include the injected types in the window object's type, one can include this package's `extend-window.d.ts` file when building.
-This can be acheived by adding the following to your project's `tsconfig.json` file:
-
-```json
-{
-    ...
-    "include": [..., "path/to/node_modules/@concordium/browser-wallet-api-types/extend-window.d.ts"]
-}
-```
-
```
soerenbf, line 68: Is this section not needed anymore, or why has this been removed? It's stated that "the package changes the type of the window object", but it doesn't change the type until you include this file (or its equivalent in the `/lib` folder).

I realize we want to push using the detector to get the api, but people might still be interested in this anyway?

packages/browser-wallet-api-helpers/webpack.config.js
```diff
@@ -0,0 +1,50 @@
...
+                        plugins: ['@babel/plugin-transform-runtime', '@babel/plugin-transform-modules-commonjs'],
+                    },
+                },
+            },
+        ],
+    },
+    output: {
+        filename: '[name].min.js',
+        path: path.resolve(__dirname, 'lib'),
+        library: 'concordiumApiHelpers',
```
soerenbf, line 46: Maybe `concordiumHelpers`, as the API injected is not accessible on `concordiumApi`, but just `concordium`?

packages/browser-wallet-api-helpers/README.md
orhoj: Søren and I discussed it yesterday and the conclusion was that advanced users would figure this out themselves. For MetaMask people define their own global window in their application when accessing it from window directly. They do not include it like this. 

But I should update the documentation.

packages/browser-wallet-api-helpers/webpack.config.js
orhoj: Works for me.

packages/browser-wallet-api-helpers/README.md
orhoj: I added a description for how to use `window.concordium` directly.

packages/browser-wallet-api-helpers/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@concordium/browser-wallet-api-helpers",
+    "version": "0.1.0",
+    "license": "Apache-2.0",
+    "packageManager": "yarn@3.2.0",
+    "main": "lib/index.js",
```
shjortConcordium, line 6:
```suggestion
    "main": "lib/index.js",
    "browser": "lib/concordiumHelpers.min.js"
```
At least is what I do in the web-sdk :shrimp:

shjortConcordium (approved):

PR #34, opened 2022-06-28
Account actions bar

soerenbf:
## Purpose

Adds a navigation bar to the bottom of the account, letting the user navigate to different account sections (send ccd, transaction log, etc.)

## Changes

- Added Account actions component.
- Added routes scoped to account page, which is what the navigation bar uses to change the account page content.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/page-layouts/MainLayout/Header/Header.tsx
```diff
@@ -129,7 +129,7 @@ export default function Header({ onToggle }: Props) {
                     {isHomePage || (
```
shjortConcordium, line 129: According to the design, it seems that the different account pages should have exit buttons within themselves, but that this should not appear?

soerenbf: Yes, you're right!

packages/browser-wallet/src/popup/pages/Account/AccountActions/AccountActions.tsx
```diff
@@ -49,11 +51,13 @@ function ActionLinks({ children }: ActionLinksProps) {
...
 
-export default function AccountActions() {
+type Props = ClassName;
+
+export default function AccountActions({ className }: Props) {
     const { t } = useTranslation('account', { keyPrefix: 'actions' });
 
     return (
-        <nav className="account-page-actions">
+        <nav className={clsx('account-page-actions', className)}>
```
shjortConcordium, line 60: I don't really understand why you need this className parameter (which you use to specify the flex), instead of just using `account-page-actions` for that. (Also is it not only going to used in this context?

shjortConcordium (approved):

packages/browser-wallet/src/popup/pages/Account/AccountActions/AccountActions.tsx
soerenbf: It's to specify `height` in the most recent commit, and it's because the account page uses the heights of the different elements on page to ultimately calculate the height of the scrollable area on the page (f.x. transaction log). Having all the different heights in the same css file makes it easier to understand what's going on.

PR #43, opened 2022-08-03
Account balance

orhoj:
## Purpose
Get the account balances periodically from the JSON-RPC server to show in the UI.

## Changes
- Added an event emitter where one can subscribe to getting the account info periodically. It also allows for listening to changes in the balance only.
- Errors are not handled as we have yet to define how this is to be shown in the UI. TODO's have been left in their place.
- The JSON-RPC url in the state is made mandatory, i.e. it can no longer be undefined.

## Checklist
      hard-to-understand areas.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,10 +18,12 @@ module.exports = {
     rules: {
         'import/prefer-default-export': 0,
         'no-restricted-exports': 0,
+        'no-restricted-syntax': ['error', 'ForInStatement', 'LabeledStatement', 'WithStatement'],
         'react/jsx-props-no-spreading': 0,
         'react/require-default-props': 0,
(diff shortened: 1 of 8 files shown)


shjortConcordium (approved):
LGTM, for now, though it would be nice if we added a "middleware" for the account info, because currently, when changing account, it gets the accountInfo both when you open the list and when you enter open the view for the chosen account.

orhoj:
> LGTM, for now, though it would be nice if we added a "middleware" for the account info, because currently, when changing account, it gets the accountInfo both when you open the list and when you enter open the view for the chosen account.

Yes, I think we should make this optimization at a later point. That way we can also display the latest seen value instead of defaulting to 0.

PR #44, opened 2022-08-03
Add piggybank with v1 contract + Support sign/send of V1 contract parameters

shjortConcordium:
## Purpose

Support V1 contracts + add example for them.

## Changes

- Create an example using a V1 piggybank
- Allow dApps to send schema version when sending contract transactions for signing.

## Checklist

      hard-to-understand areas.


examples/piggybank/src/Version1.tsx
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable no-console */
```
orhoj, line 1: I think there could be better reuse between Version0 and Version1.

orhoj: But we can defer that to later, as I believe there are more important things to work on right now.

orhoj (approved):

PR #46, opened 2022-08-05
Allow sending JsonRpcRequests through wallet api

shjortConcordium:
## Purpose

Expose JSON-RPC methods through wallet-api.

## Changes

- Add a JSON-RPC client to the wallet-api.
- Updated piggybank to use wallet's JSON request.

## Checklist

      hard-to-understand areas.


packages/browser-wallet-api-helpers/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.1.2
```
orhoj, line 3: This should be 0.2.0

packages/browser-wallet-api-helpers/README.md
```diff
@@ -114,3 +114,12 @@ const provider = await detectConcordiumProvider();
...
 provider.addChangeAccountListener((address) => (selectedAccountAddress = address));
 ```
+
+## Accessing node through JSON-RPC
+
+The wallet api exposes access to a JSON-RPC client. This allows a dApp to communicate with a node using JSON-RPC without knowing a server itself. The client can be accessed using the `node` field.
+
+```typescript
+const provider = await detectConcordiumProvider();
+const accountInfo = await provider.node.getAccountInfo(accountAddress);
```
orhoj, line 124: Accessing a class field directly seems a bit yucky. Perhaps it should be:
```
const accountInfo = await provider.getJsonRpcClient().getAccountInfo(accountAddress);
```

packages/browser-wallet-api-helpers/README.md
```diff
@@ -114,3 +114,12 @@ const provider = await detectConcordiumProvider();
 let selectedAccountAddress: string | undefined = undefined;
 provider.addChangeAccountListener((address) => (selectedAccountAddress = address));
 ```
+
+## Accessing node through JSON-RPC
+
+The wallet api exposes access to a JSON-RPC client. This allows a dApp to communicate with a node using JSON-RPC without knowing a server itself. The client can be accessed using the `node` field.
```
orhoj, line 120:
```suggestion
The wallet API exposes access to a JSON-RPC client. This allows a dApp to communicate with the same node as the wallet is connected to, and enables dApps to access the JSON-RPC interface without being connected to a separate server itself. The client is accessed as shown in the example below.
```

packages/browser-wallet/src/background/index.ts
```diff
@@ -4,11 +4,44 @@ import {
...
     MessageType,
 } from '@concordium/browser-wallet-message-hub';
-import { storedSelectedAccount, storedUrlWhitelist } from '@shared/storage/access';
+import { storedSelectedAccount, storedUrlWhitelist, storedJsonRpcUrl } from '@shared/storage/access';
 
+import { v4 as uuidv4 } from 'uuid';
 import bgMessageHandler from './message-handler';
 import { forwardToPopup, HandleMessage, HandleResponse, RunCondition, setPopupSize } from './window-management';
 
+// TODO Replace with concordiumSDK HttpProvider when cross-fetch supports service workers. https://github.com/lquixada/cross-fetch/issues/78
```
orhoj, line 13: Could the HttpProvider be refactored to work in a service worker?

packages/browser-wallet/src/background/index.ts
```diff
@@ -4,11 +4,44 @@ import {
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const jsonRpcRequest = async (url: string, method: string, params?: any) => {
```
orhoj, line 15: Is `params` a `string`?

shjortConcordium: Removed this.

packages/browser-wallet/src/background/index.ts
shjortConcordium: https://github.com/Concordium/concordium-node-sdk-js/pull/82

packages/browser-wallet-api-helpers/README.md
```diff
@@ -147,3 +147,12 @@ provider.on('accountDisconnected', (accountAddress) => {
...
 provider.connect().then((accountAddress) => (selectedAccountAddress = accountAddress));
 ```
+
+## Accessing node through JSON-RPC
+
+The wallet API exposes access to a JSON-RPC client. This allows a dApp to communicate with the same node as the wallet is connected to, and enables dApps to access the JSON-RPC interface without being connected to a separate server itself. The client is accessed as shown in the example below.
+
+```typescript
+const provider = await detectConcordiumProvider();
+const accountInfo = await provider.getJsonRpcClient().getAccountInfo(accountAddress);
```
orhoj, line 157: I think the example should show how to get the client, and then it should also state that the client will be outdated if the network changes.

packages/browser-wallet/build/dev.ts
```diff
@@ -32,6 +32,7 @@ const config: BuildOptions = {
     outdir: 'dist',
     define: {
         'process.env.NODE_ENV': `"${process.env.NODE_ENV}"`,
+        'document.baseURI': '""',
```
orhoj, line 35: This magic should be documented, or even better if we didn't need it.

packages/browser-wallet-api/src/wallet-api.ts
```diff
@@ -92,6 +104,10 @@ class WalletApi extends EventEmitter implements IWalletApi {
     private handleEvent(type: EventType) {
         this.messageHandler.handleMessage(createEventTypeFilter(type), (msg) => this.emit(type, msg.payload));
     }
+
+    public getJsonRpcClient(): JsonRpcClient {
```
orhoj, line 108: Is there access restriction on this? A non-whitelisted website should not be able to use it.

orhoj:
If I don't have any account connected and I try to get the JSON-RPC client, then a connection request in the wallet is opened.

orhoj:
The returned errors from the client should be better. I just get that the RPC call was rejected by the wallet, but my request was in fact just bad.

packages/browser-wallet-api-helpers/README.md
```diff
@@ -147,3 +151,18 @@ provider.on('accountDisconnected', (accountAddress) => {
 // Connect to an account in the wallet again triggered elsewhere in the dApp.
 provider.connect().then((accountAddress) => (selectedAccountAddress = accountAddress));
 ```
+
+## Accessing node through JSON-RPC
+
+The wallet API exposes access to a JSON-RPC client. This allows a dApp to communicate with the same node as the wallet is connected to, and enables dApps to access the JSON-RPC interface without being connected to a separate server itself. The client is accessed as shown in the example below.
+The dApp does not need to recreate the client again when the wallet changes node or network, the client will always use the wallet's current connected JSON-RPC server.
+
+If you have not connected with the wallet (or previously been whitelisted) or if the user rejects signing the transaction, the commands will not be executed and the method will throw an error.
```
orhoj, line 160: If the user rejects signing the transactions? That does not seem relevant here.

shjortConcordium: Yes, it is not. Removed it.

packages/browser-wallet/src/background/index.ts
```diff
@@ -4,11 +4,54 @@ import {
...
+                    // We lose the method's typing when sending the message.
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    method as any,
+                    params && JSONBig.parse(params)
+                )
+                .then(onSuccess)
+                .catch((e) => onFailure(e.toString()));
+        }
+    } else {
+        onFailure('RPC Call can only performed by whitelisted sites');
```
orhoj, line 51:
```suggestion
        onFailure('RPC Call can only be performed by whitelisted sites');
```

orhoj (approved):

PR #47, opened 2022-08-10
Connection setup

orhoj:
## Purpose
Make whitelisting of dApps a per-account setting instead of a global setting.

## Changes
- Added an account settings page.
- Added a connected sites page for handling the dApps that an account is connected to. Currently it only allows for disconnecting from a site.
- When a dApp connects we connect the currently selected account. Later this should/could be expanded to provide a list of accounts to connect to.
- signMessage and signTransaction now require dApps to provide the account that should perform the action, as it no longer defaults to the selected account.
- the connect() API will return an address of a connected account if any is present, not only if the selected account is connected. The selected account is still prioritized as the first value to return if connected.
- Removed url whitelist storage as it has been replaced by the connected sites.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/piggybank/src/Root.tsx
+++ b/examples/piggybank/src/Root.tsx
@@ -58,15 +58,15 @@ const state = createContext<State>({ isConnected: false, account: undefined, jso
 /**
  * Action for depositing an amount of microCCD to the piggy bank instance
  */
-const deposit = (amount = 0) => {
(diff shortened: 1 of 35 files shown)


orhoj:
Update the documentation in browser-wallet-api-helpers as the API's have changed.

shjortConcordium (approved):
Yes, please update api-helpers CHANGELOG and README.

PR #48, opened 2022-08-11
Add ability to manually add connection to a site

orhoj:
## Purpose
Allow users to manually connect to a site. Broadcast a disconnect event when removing a connection and an accountChanged event when adding a new connection, but only to the given site.

## Changes
- Added new disconnect event to WalletApi. Emitted to the disconnected tab.
- Added option to manually connect the currently open tab's website to the wallet.
- An accountChanged event is emitted when manually connecting to a site, but only to that site.
- Connected site whitelisting now uses the `origin` field on the URL instead of the full URL which was too restrictive.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Account/AccountSettings/ConnectedSites.tsx
+++ b/packages/browser-wallet/src/popup/pages/Account/AccountSettings/ConnectedSites.tsx
@@ -1,21 +1,37 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { useAtom, useAtomValue } from 'jotai';
 import { selectedAccountAtom, storedConnectedSitesAtom } from '@popup/store/account';
 import Button from '@popup/shared/Button';
 import { useTranslation } from 'react-i18next';
+import { popupMessageHandler } from '@popup/shared/message-handler';
+import { EventType } from '@concordium/browser-wallet-api-helpers';
 
 export default function ConnectedSites() {
     const { t } = useTranslation('account', { keyPrefix: 'settings.connectedSites' });
     const selectedAccount = useAtomValue(selectedAccountAtom);
     const [connectedSitesLoading, setConnectedSites] = useAtom(storedConnectedSitesAtom);
     const connectedSites = connectedSitesLoading.value;
+    const [openTabUrl, setOpenTabUrl] = useState<string>();
+
+    async function getCurrentOpenTabUrl() {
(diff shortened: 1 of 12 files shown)


packages/browser-wallet/src/popup/pages/Account/AccountSettings/ConnectedSites.tsx
```diff
@@ -1,46 +1,99 @@
...
+import { EventType } from '@concordium/browser-wallet-api-helpers';
 
 export default function ConnectedSites() {
     const { t } = useTranslation('account', { keyPrefix: 'settings.connectedSites' });
     const selectedAccount = useAtomValue(selectedAccountAtom);
     const [connectedSitesLoading, setConnectedSites] = useAtom(storedConnectedSitesAtom);
     const connectedSites = connectedSitesLoading.value;
+    const [openTabUrl, setOpenTabUrl] = useState<string>();
+
+    async function getCurrentOpenTabUrl() {
```
shjortConcordium, line 16: Move this out of the component.

packages/browser-wallet/src/popup/pages/Account/AccountSettings/ConnectedSites.tsx
```diff
@@ -1,46 +1,99 @@
...
+        const tabs = await chrome.tabs.query({ active: true, lastFocusedWindow: true });
+        const { url } = tabs[0];
+        if (!url) {
+            return undefined;
+        }
+        return new URL(url).origin;
```
shjortConcordium, line 22: The origin differentiates between http and https, but maybe that is preferable?

packages/browser-wallet/src/popup/pages/Account/AccountSettings/ConnectedSites.tsx
orhoj: I have done that in the following PR, so I will leave it be here to make merging easier: https://github.com/Concordium/concordium-browser-wallet/blob/3155f9ead7985a02db082ef5846caf44cdb1fc76/packages/browser-wallet/src/popup/shared/utils/tabs.ts.

packages/browser-wallet/src/popup/pages/Account/AccountSettings/ConnectedSites.tsx
orhoj: We will keep it as-is as that is more restrictive.

shjortConcordium (approved):

PR #59, opened 2022-08-23
Confirm accounts and identities

shjortConcordium:
## Purpose

Confirm accounts and identities in the background of the wallet.

## Changes

- Move confirmation of identities to the background script, and check globally instead of only in the identity view.
- Add confirmation of accounts.
- Move creation of accounts to background script.

## Checklist

      hard-to-understand areas.


orhoj:
After my account finalization the account view was not updated.

orhoj:
The "Your Concordium identity" popup is empty.

orhoj:
A pending identity that is rejected or verified does not update in the ID Cards view. If I navigate away and back, then it does update to the new value.

packages/browser-wallet/src/background/startup.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+        setTimeout(loop, updateInterval);
+    }, 0);
+}
+
+/**
+ * Continously checks whether pending identities have been confirmed or rejected.
+ */
+async function monitorIdentityStatus() {
+    setTimeout(async function loop() {
```
orhoj, line 55: This does not seem like an ideal solution. We just have an infinite loop going on, instead of just checking the status on-demand. Can you make it so that the front-end pushes a message that ask for listening to the status instead? The only at startup do we need to read all stored identities to start listening for the pending ones, and the UI will then ask when it creates a new identity.

packages/browser-wallet/src/background/startup.ts
```diff
@@ -0,0 +1,87 @@
...
+        const identities = await storedIdentities.get();
+        const toUpdate: Identity[] = [];
+        if (identities) {
+            for (const { location, ...identity } of identities.filter(isPendingIdentity)) {
+                const response = (await (await fetch(location)).json()) as IdentityTokenContainer;
```
orhoj, line 60: If this fails, then the whole loop just stops.

packages/browser-wallet/src/background/startup.ts
orhoj: If it seems like a bigger thing to refactor, then I can live with the loops initially and then we can re-do it later.

packages/browser-wallet/src/background/startup.ts
```diff
@@ -0,0 +1,87 @@
...
+    setTimeout(async function loop() {
+        const network = await storedCurrentNetwork.get();
+        const url = network?.jsonRpcUrl;
+        const creds = await storedCredentials.get();
+        const toUpdate: WalletCredential[] = [];
+        if (url && creds) {
+            const client = new JsonRpcClient(new HttpProvider(url, fetch));
+            for (const { status, deploymentHash, ...info } of creds.filter(isPendingCred)) {
+                const response = await client.getTransactionStatus(deploymentHash);
+                if (response?.status === 'finalized') {
```
orhoj, line 33:
```suggestion
                if (response?.status === TransactionStatusEnum.Finalized) {
```

packages/browser-wallet/src/background/update.ts
```diff
@@ -0,0 +1,67 @@
...
+async function editList<Type>(
+    lock: string,
+    edits: Type[],
+    findPredicate: (current: Type) => (candidate: Type) => boolean,
+    storage: StorageAccessor<Type[]>
+): Promise<void> {
+    return navigator.locks.request(lock, async () => {
+        const currentList = await storage.get();
+        if (!currentList) {
+            return;
```
orhoj, line 30: It may be a better idea to throw an error in this case. I am trying to edit a non-existing list and the method just goes well, so I am now under the assumption that my edit went through, but nothing happened.

orhoj (changes requested):

packages/browser-wallet/src/background/startup.ts
shjortConcordium: I'll fix it later.

packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -80,3 +83,19 @@ export function atomWithChromeStorage<V>(key: ChromeStorageKey, fallback: V, wit
 
     return derived;
 }
+
+export function useListenForUpdates<V>(key: ChromeStorageKey, atomToUpdate: WritableAtom<V, V, void>) {
```
orhoj, line 87: There might be some infinite looping possible with this. I suggest that we improve our atoms with a way to make them listen to changes to the storage.

shjortConcordium: I did this :tada:

packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -68,6 +71,21 @@ export function atomWithChromeStorage<V>(key: ChromeStorageKey, fallback: V, wit
...
+            const listener = (changes: Record<string, chrome.storage.StorageChange>) =>
+                getStoredValue().then((value) => {
+                    if (key in changes) {
+                        setValue({
+                            loading: false,
+                            value: value ?? fallback,
+                        });
+                    }
+                });
+            chrome.storage.local.onChanged.addListener(listener);
```
orhoj, line 85: `local` could also be any other chrome storage type, so it should be made more generic.

shjortConcordium: Yes, I have fixed that now.

orhoj (approved):

PR #68, opened 2022-08-26
Indexed storage

shjortConcordium:
## Purpose

 - Index storage of identities/accounts/IDPs under the current network.
 - Reset selected identity/account when switching network.
 - Block changing network during identity issuance.

## Changes

The purpose +
 - Removed identity id, and net from identities/credentials

## Checklist

      hard-to-understand areas.


orhoj:
The list of identity providers is not reset when switching network. It probably works because we update it on-demand, but perhaps it should then be cleared when switching network to avoid any cross-polution.

orhoj:
When switching network we should default a selected account. For example I only have 1 account, and it might as well become the selected account when switching around instead of showing a blank account screen.

packages/browser-wallet/src/popup/store/settings.ts
```diff
@@ -22,9 +23,13 @@ const storedNetworkConfigurationAtom = atomWithChromeStorage<NetworkConfiguratio
 );
 export const networkConfigurationAtom = atom<NetworkConfiguration, NetworkConfiguration>(
     (get) => get(storedNetworkConfigurationAtom),
-    (_, set, networkConfiguration) => {
+    (get, set, networkConfiguration) => {
         set(storedNetworkConfigurationAtom, networkConfiguration);
         popupMessageHandler.broadcast(EventType.ChainChanged, networkConfiguration.genesisHash);
+        const accounts = get(accountsAtom);
+        const identities = get(identitiesAtom);
+        set(selectedAccountAtom, accounts.length ? accounts[0] : undefined);
```
orhoj, line 31: I guess my previous comment is that this does not seem to work.

packages/browser-wallet/src/popup/pages/NetworkSettings/NetworkSettings.tsx
```diff
@@ -78,8 +78,34 @@ function NetworkConfigurationComponent({ networkConfiguration }: { networkConfig
...
+            <Modal disableClose open={Boolean(pendingIdentity)}>
+                <div className="network-settings-page__pending-identity">
+                    <p>{t('pendingIdentity.description')}</p>
+                    <Button
+                        width="narrow"
+                        className="network-settings-page__pending-identity__button"
+                        onClick={() => nav(-1)}
+                    >
+                        {t('pendingIdentity.back')}
+                        go back
```
orhoj, line 96: Translate or remove?

packages/browser-wallet/src/popup/pages/NetworkSettings/NetworkSettings.tsx
```diff
@@ -78,8 +78,34 @@ function NetworkConfigurationComponent({ networkConfiguration }: { networkConfig
...
+                    </Button>
+                    <Button
+                        faded
+                        width="narrow"
+                        className="network-settings-page__pending-identity__button"
+                        onClick={() => setPendingIdentity(undefined)}
+                    >
+                        {t('pendingIdentity.abort')}
+                        abort issuance
```
orhoj, line 105: Translate or remove?

orhoj (changes requested):

packages/browser-wallet/src/background/startup.ts
```diff
@@ -23,26 +23,28 @@ async function monitorAccountStatus() {
...
-                        if (response?.status === TransactionStatusEnum.Finalized) {
-                            const isSuccessful = Object.values(response?.outcomes || {}).some(
-                                (outcome) => outcome.result.outcome === 'success'
-                            );
-                            toUpdate.push({
-                                ...info,
-                                status: isSuccessful ? CreationStatus.Confirmed : CreationStatus.Rejected,
-                            });
+            if (network) {
+                const url = network?.jsonRpcUrl;
```
orhoj, line 27:
```suggestion
                const url = network.jsonRpcUrl;
```

packages/browser-wallet/src/shared/storage/access.ts
```diff
@@ -44,28 +63,77 @@ const makeStorageAccessor = <V>(area: chrome.storage.AreaName, key: ChromeStorag
...
+    area: chrome.storage.AreaName,
+    key: ChromeStorageKey
+): IndexedStorageAccessor<Value> => {
+    const inner = makeStorageAccessor<Record<string, Value>>(area, key);
+    return {
+        get: (index: string) => inner.get().then((s) => (s ? s[index] : undefined)),
+        set: async (index: string, value: Value) => {
+            const current = await inner.get();
+            const updated = { ...current };
+            updated[index] = value;
```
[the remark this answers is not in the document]
shjortConcordium: I think that is the same :thinking:

packages/browser-wallet/src/popup/store/settings.ts
shjortConcordium: Yeah, it used the list from the previous net. I've changed this, so it uses the correct net.

packages/browser-wallet/src/shared/storage/access.ts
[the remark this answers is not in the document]
shjortConcordium: or at least this has the same behaviour as the normal accessor.

orhoj (approved):

PR #75, opened 2022-09-05
Update log on balance change

orhoj:
## Purpose
Ensure that the transaction history is updated when the balance for an account changes, i.e. if a new transfer is received.

## Changes
- Update transaction history if the account balance changes.
- Use chrome session storage to cache account info to optimize how often we fetch the account info, and also to use the latest seen value in account views instead of showing 0.

## Checklist
      hard-to-understand areas.


packages/browser-wallet/src/popup/page-layouts/MainLayout/AccountList/AccountList.tsx
```diff
@@ -58,24 +56,17 @@ const AccountList = forwardRef<HTMLDivElement, Props>(({ className, onSelect },
...
-            emitter.removeAllListeners('totalchanged');
-            emitter.stop();
-        };
-    }, [JSON.stringify(accounts)]);
+        for (const info of accountInfos) {
+            if (info) {
+                setTotalBalanceMap(new Map(totalBalanceMap.set(info.accountAddress, info.accountAmount)));
+            }
+        }
+    }, [JSONBig.stringify(accountInfos)]);
```
shjortConcordium, line 69: This could be refactored to only `setTotalBalanceMap` once. (I think that might be better :thinking:)

packages/browser-wallet/src/popup/shared/account-info-emitter.ts
```diff
@@ -31,43 +24,47 @@ export class AccountInfoEmitter extends EventEmitter {
...
-                        this.emit('accountinfo', accountInfo, address.address);
-                        if (accountInfo.accountAmount !== this.previousTotalMap.get(address.address)) {
-                            this.emit('totalchanged', accountInfo, address.address);
-                            this.previousTotalMap.set(address.address, accountInfo.accountAmount);
+                if (this.accounts.length > 0) {
+                    const lfBlockHash = (await this.client.getConsensusStatus()).lastFinalizedBlock;
+                    for (const address of this.accounts) {
+                        const accountInfo = await this.client.getAccountInfo(address, lfBlockHash);
+                        if (accountInfo) {
+                            const network = await storedCurrentNetwork.get();
```
shjortConcordium, line 58: Maybe we can avoid reading this from memory for each account?

packages/browser-wallet/src/popup/shared/AccountInfoEmitterContext/AccountInfoEmitterContext.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const { genesisHash } = useAtomValue(networkConfigurationAtom);
+    const address = useMemo(() => account.address, [account]);
+    const addToast = useSetAtom(addToastAtom);
+    const { jsonRpcUrl } = useAtomValue(networkConfigurationAtom);
+    const { t } = useTranslation();
+
+    const [accountInfo, setAccountInfo] = useState<AccountInfo>();
+
+    useEffect(() => {
+        if (accountInfoCache && accountInfoCache[address]) {
```
shjortConcordium, line 68:
```suggestion
        if (accountInfoCache[address]) {
```
Otherwise you need to be more "defensive" in the dep array.

packages/browser-wallet/src/popup/shared/AccountInfoEmitterContext/AccountInfoEmitterContext.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }, [genesisHash, address, accountInfoCache[address]]);
+
+    useEffect(() => {
+        if (account.status === CreationStatus.Confirmed && accountInfoEmitter) {
+            const listener = accountInfoEmitter.subscribe(address, noOp);
+            return () => {
+                accountInfoEmitter.unsubscribe(address, listener);
+            };
+        }
+        return () => noOp;
```
shjortConcordium, line 100:
```suggestion
        return noOp;
```

packages/browser-wallet/src/popup/shared/account-info-emitter.ts
```diff
@@ -31,43 +24,47 @@ export class AccountInfoEmitter extends EventEmitter {
...
-            this.emit('error', e);
+    unsubscribe(
+        accountAddress: string,
+        listener: (accountInfo: AccountInfo) => void
+    ): (accountInfo: AccountInfo) => void {
+        this.removeListener(accountAddress, listener);
+        if (this.listenerCount(accountAddress) === 0) {
+            this.accounts = this.accounts.filter((address) => accountAddress !== address.address);
         }
+        return listener;
```
shjortConcordium, line 47: Why do we return the listener here?
(I kinda see the niceness of returning it with subscribe, but why here?)

packages/browser-wallet/tsconfig.json
```diff
@@ -21,5 +21,7 @@
             "module": "CommonJS"
         }
     },
-    "include": ["./**/*", "../../node_modules/@concordium/browser-wallet-api-helpers/lib/extend-window.d.ts"]
+    "include": ["./**/*"],
+    "exclude": ["./../../node_modules/**/*", "./dist/**/*"],
+    "files": ["../../node_modules/@concordium/browser-wallet-api-helpers/lib/extend-window.d.ts"]
```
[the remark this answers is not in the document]
orhoj: Yes. If added to `files`, then the `exclude` does not overrule it.

packages/browser-wallet/src/popup/shared/account-info-emitter.ts
orhoj: Actually this value was unused as we use `getGenesisHash` for the indexed storage. I have, however, moved that out of the loop to avoid this issue.

packages/browser-wallet/src/popup/shared/account-info-emitter.ts
orhoj: For no reason.

packages/browser-wallet/src/popup/shared/AccountInfoEmitterContext/AccountInfoEmitterContext.tsx
```diff
@@ -0,0 +1,104 @@
...
+                        })
+                        .catch(() => addToast(t('account.error')));
+                })
+                .catch(() => addToast(t('account.error')));
+        }
+    }, [genesisHash, address, accountInfoCache[address]]);
+
+    useEffect(() => {
+        if (account.status === CreationStatus.Confirmed && accountInfoEmitter) {
+            const listener = accountInfoEmitter.subscribe(address, noOp);
```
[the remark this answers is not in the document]
orhoj: No, I kind of just kept it around if we wanted to use it at some point, but we probably won't. I guess it's only going to be the error events that we need a listener for. I will clean it up, and then we can add stuff back if ever need it.

packages/browser-wallet/src/popup/shared/AccountInfoListenerContext/AccountInfoListenerContext.tsx
```diff
@@ -88,16 +89,16 @@ export function useAccountInfo(account: WalletCredential): AccountInfo | undefin
                 })
                 .catch(() => addToast(t('account.error')));
         }
-    }, [genesisHash, address, accountInfoCache[address]]);
+    }, [genesisHash, address, accountInfoCache, accountInfoCache[address]]);
```
shjortConcordium, line 92: I'm not happy about this.

shjortConcordium (approved):

packages/browser-wallet/src/popup/shared/AccountInfoListenerContext/AccountInfoListenerContext.tsx
orhoj: I did like you wanted it.

PR #77, opened 2022-09-05
Faucet

orhoj:
## Purpose
Add a testnet/stagenet faucet button for accounts that have 0 transactions.

## Changes
- Added a faucet button when connected to a non-mainnet network, for accounts with 0 transactions.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
```diff
@@ -140,6 +143,44 @@ function InfiniteTransactionList({
...
 
+/**
+ * Update an existing list of descending transactions with a new list of ascending transactions,
+ * ensuring that any overlaps are updated with the values from the new list of transactions.
+ */
+function updateTransactionsWithNewTransactions(
+    existingTransactions: BrowserWalletTransaction[],
+    newTransactions: BrowserWalletTransaction[]
+) {
+    const existingHashes = existingTransactions.map((trs) => trs.transactionHash);
```
shjortConcordium, line 154: Please filter out rewards/undefined hashes here so that the rewards don't update each other.

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
```diff
@@ -140,6 +143,44 @@ function InfiniteTransactionList({
...
+    const transactionUpdates: Record<string, BrowserWalletTransaction> = {} as Record<string, BrowserWalletTransaction>;
+
+    // Collect updates to existing transactions.
+    newTransactions
```
shjortConcordium, line 158: You could do a split instead of a filter here, so you can avoid going through the list again later. :thinking:

orhoj: I have used the `partition` function instead.

PR #81, opened 2022-09-07
Recover accounts

shjortConcordium:
## Purpose

Also recover accounts.

## Changes

Recover accounts
Update UI in recovery page.

## Checklist

      hard-to-understand areas.


orhoj:
The `Continue` button after having recovered is not placed correctly on the screen.

orhoj:
After having recovered an account, there is not selected account in the account view. I can select the account, but it should be selected automatically after recovery.

orhoj:
On failure the `Retry` button should be placed at the bottom like the `Continue` button.

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -1,23 +1,71 @@
...
+import { storedCurrentNetwork } from '@shared/storage/access';
+import { addCredential, addIdentity } from './update';
 
 // TODO: increase max empty
 // How many empty identityIndices are allowed before stopping
 const maxEmpty = 3;
 // Milliseconds to wait, if retrievalUrl is still pending
 const sleepInterval = 5000;
 
+function getAddressFrom(accountInfo: AccountInfo) {
```
orhoj, line 29: You can do:
```
accountInfo.accountAddress
```

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -1,23 +1,71 @@
...
-import { Identity, CreationStatus, IdentityProvider } from '@shared/storage/types';
 import { BackgroundResponseStatus } from '@shared/utils/types';
-import { addIdentity } from './update';
+import { Identity, CreationStatus, IdentityProvider, WalletCredential } from '@shared/storage/types';
+import { storedCurrentNetwork } from '@shared/storage/access';
+import { addCredential, addIdentity } from './update';
 
 // TODO: increase max empty
 // How many empty identityIndices are allowed before stopping
 const maxEmpty = 3;
```
orhoj, line 25: Update this to be 10, and remove the TODO.

orhoj (changes requested):

packages/browser-wallet/src/background/recovery.ts
shjortConcordium: Because 10 is a good sample size?

orhoj: Because I asked Lars to also use 10.

orhoj:
The recovery screen CSS is messed up.

shjortConcordium:
@orhoj I fixed the ui + updated the identity recovery to work with the new v1 version.

PR #82, opened 2022-09-08
Block duplicate account issue + fix other bugs

shjortConcordium:
## Purpose

Fix duplicate account issue + crashing when creating account + error appearing in log in setup.

## Changes

+ Fixed `makeUncontrolled` valid prop.
+ AccountList no longer handles account infos, this has been pushed to list items.
+ Stop user from starting another credential request.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/state/index.ts
+++ b/packages/browser-wallet/src/popup/state/index.ts
@@ -10,4 +10,3 @@ export const addToastAtom = atom<null, string>(null, (get, set, newToast) => {
     const updatedToasts = [...currentToasts, newToast];
     set(toastsAtom, updatedToasts);
 });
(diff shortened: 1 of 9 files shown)


packages/browser-wallet/src/popup/state/index.ts
```diff
@@ -10,3 +10,4 @@ export const addToastAtom = atom<null, string>(null, (get, set, newToast) => {
     const updatedToasts = [...currentToasts, newToast];
     set(toastsAtom, updatedToasts);
 });
+export const creatingCredentialRequestAtom = atom<boolean>(false);
```
orhoj, line 13: This has to be saved in session storage, otherwise I can just close the popup and open it again and then get access to creating duplicate accounts.

PR #84, opened 2022-09-09
Add recovery from settings

shjortConcordium:
## Purpose

Add recovery from settings.

## Changes

Add page under settings for recovery.
Moved logic from `PerformRecovery` to Recovery Page.

## Checklist

      hard-to-understand areas.


orhoj:
When done recovering it says it has recovered the identity and account I already have in the wallet.

packages/browser-wallet/src/popup/pages/IdentityIssuance/IdentityIssuanceEnd.tsx
```diff
@@ -25,18 +25,14 @@ interface Location {
...
 }
 
-interface Props {
-    onFinish: () => void;
-}
-
 // TODO start checking status on identity
 
-export default function IdentityIssuanceEnd({ onFinish }: Props) {
+export default function IdentityIssuanceEnd() {
```
orhoj, line 30: This component should have some padding around the text that matches what we use elsewhere. Currently it touches the boundaries of the window.

packages/browser-wallet/src/popup/pages/Recovery/RecoveryFinish.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export default function DisplayRecoveryResult() {
+    const { state } = useLocation() as Location;
+    const { payload } = state;
+    const { setReturnLocation, withClose } = useContext(fullscreenPromptContext);
+    const { t } = useTranslation('recovery');
+    const navigate = useNavigate();
+    const identities = useAtomValue(identitiesAtom);
+    const credentials = useAtomValue(credentialsAtom);
```
orhoj, line 28: You can't just show everything in the wallet as being what has been recovered. That only works for an empty wallet.

packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { identityProvidersAtom, isRecoveringAtom } from '@popup/store/identity';
+import PendingArrows from '@assets/svg/pending-arrows.svg';
+import { BackgroundResponseStatus } from '@shared/utils/types';
+import { getIdentityProviders } from '@popup/shared/utils/wallet-proxy';
+import { getNet } from '@shared/utils/network-helpers';
+import { NetworkConfiguration } from '@shared/storage/types';
+import PageHeader from '@popup/shared/PageHeader';
+import { absoluteRoutes } from '@popup/constants/routes';
+
+async function getGlobal(network: NetworkConfiguration): Promise<CryptographicParameters> {
```
orhoj, line 18: This seems like a function that should be shared, as we have multiple places where we get global.

packages/browser-wallet/src/popup/page-layouts/MainLayout/i18n/en.ts
```diff
@@ -8,6 +8,7 @@ const t = {
         accounts: 'Accounts',
         ids: 'ID cards',
         settings: 'Wallet settings',
+        recovery: 'Restoring your wallet',
```
orhoj, line 11: It's a little weird that this settings entry has its own header title, meanwhile the others just show Wallet settings.

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -128,11 +153,15 @@ async function performRecovery({ providers, ...recoveryInputs }: Payload) {
     }
     await addIdentity(identitiesToAdd);
     await addCredential(credsToAdd);
+    await sessionIsRecovering.set(false);
```
orhoj, line 156: If some error happens during the recovery method, then this will not be set to `false` as far as I can see. So if I open the wallet, then what happens? It will show the spinner but it will never complete?

packages/browser-wallet/src/popup/pages/Recovery/i18n/da.ts
```diff
@@ -0,0 +1,22 @@
+import type en from './en';
+
+const t: typeof en = {
+    intro: {
+        description:
+            'If you are missing some identities or accounts in your wallet, you can try restoring them here.\n\nIf you used your secret recovery phrase to create more identities and accounts in another installation of the Concordium wallet, you can also use this option to add them to this installation.',
```
orhoj, line 6: Has to be in Danish.

orhoj (changes requested):

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -76,6 +79,14 @@ async function performRecovery({ providers, ...recoveryInputs }: Payload) {
     if (!network) {
         throw new Error('No chosen network could be found');
     }
+    const identities = await storedIdentities.get(network.genesisHash);
+    const credentials = await storedCredentials.get(network.genesisHash);
+
+    const getNextCredIndex = ({ index, providerIndex }: Identity) =>
```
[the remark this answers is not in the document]
shjortConcordium: Potato, Potato

packages/browser-wallet/src/popup/page-layouts/ExternalRequestLayout/ExternalRequestLayout.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const account = useSelectedCredential();
+
+    return (
+        <>
+            <Header />
+            <div className="external-request-layout">
+                <AccountInfoListenerContext>
+                    {account && <AccountDetails expanded={false} account={account} />}
+                    <main className="external-request-layout__main">{children}</main>
+                    <Toast />
```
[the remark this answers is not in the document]
shjortConcordium: I would assume that they are in different contexts? So they would have each they own instance of the atom :thinking:
(This might need to be tested)

packages/browser-wallet/src/popup/page-layouts/MainLayout/i18n/en.ts
shjortConcordium: Yeah, I checked the figma, and it turns out they are all supposed to have their own title. (So I added that)

packages/browser-wallet/src/background/recovery.ts
shjortConcordium: Yes. :fearful:

packages/browser-wallet/src/popup/pages/Recovery/RecoveryFinish.tsx
shjortConcordium: I changed it so the recovery sends which identities/accounts have been recovered to this page, and it only displays what was recovered ( + identities that has accounts added)

packages/browser-wallet/src/popup/page-layouts/ExternalRequestLayout/ExternalRequestLayout.tsx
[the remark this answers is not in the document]
orhoj: Ah, you're probably right. I am too used to thinking about the storage syncing stuff. I am not 100% sure about the context of the atom and if they will have on each.

orhoj (approved):

PR #85, opened 2022-09-09
Update checks for creating identity and account in background script

shjortConcordium:
## Purpose

Ensure that the checks that prevents duplicate account/identity creation is set in the background script.

## Changes

Move the update of the checks from popup to background.
Set atomsWithStorage to have sync as default.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/background/credential-deployment.ts
+++ b/packages/browser-wallet/src/background/credential-deployment.ts
@@ -14,38 +14,42 @@ import { BackgroundResponseStatus, CredentialDeploymentBackgroundResponse } from
 import { addCredential } from './update';
 
 async function createAndSendCredential(credIn: CredentialInputV1): Promise<CredentialDeploymentBackgroundResponse> {
-    const network = await storedCurrentNetwork.get();
-    const url = network?.jsonRpcUrl;
-    if (!url) {
-        throw new Error('No JSON RPC url available');
+    let address: string;
+    try {
+        const network = await storedCurrentNetwork.get();
+        const url = network?.jsonRpcUrl;
+        if (!url) {
+            throw new Error('No JSON RPC url available');
+        }
+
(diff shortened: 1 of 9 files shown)


packages/browser-wallet/src/background/credential-deployment.ts
```diff
@@ -42,6 +42,10 @@ async function createAndSendCredential(credIn: CredentialInputV1): Promise<Crede
 
     // Add Pending
     await addCredential(newCred);
+    // Set Selected to new account
+    await storedSelectedAccount.set(address);
+    // Remove guard stopping another credential being created
+    await sessionCreatingCredential.set(false);
```
orhoj, line 48: I think that the removal of the guard has to be in a finally clause, e.g. if the deployment is not successful, then the guard is not reset.

packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -60,7 +60,7 @@ export function atomWithChromeStorage<V>(
  * @description
  * Create an atom that automatically syncs with chrome local storage.
  */
-export function atomWithChromeStorage<V>(key: ChromeStorageKey, fallback: V, withLoading = false, withSync = false) {
+export function atomWithChromeStorage<V>(key: ChromeStorageKey, fallback: V, withLoading = false, withSync = true) {
```
orhoj, line 63: I guess this is the quick-and-dirty way of doing it, but I guess the parameter could be completely removed instead. Or is it because you are uncertain if we might not want it as the default at some point?

shjortConcordium: Yeah I removed it completely. I thought maybe we wanted to let the atom not have sync. But we can add that option back if we need it, I guess. :shrug:

orhoj (approved):

PR #88, opened 2022-09-12
Prevent copy and paste of seed phrase in setup

shjortConcordium:
## Purpose

Disable copy and paste of seed phrase during setup.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Setup/GenerateSeedPhrase.tsx
+++ b/packages/browser-wallet/src/popup/pages/Setup/GenerateSeedPhrase.tsx
@@ -28,7 +28,14 @@ export default function GenerateSeedPhrase() {
             <div className="onboarding-setup__page-with-header">
                 <div className="onboarding-setup__page-with-header__description">{t('recoveryPhrase.description')}</div>
                 <div className="p-10">
-                    <ControlledTextArea value={seedPhrase} readOnly />
+                    <ControlledTextArea
+                        value={seedPhrase}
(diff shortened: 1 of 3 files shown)


packages/browser-wallet/src/popup/pages/Setup/GenerateSeedPhrase.tsx
```diff
@@ -28,7 +28,14 @@ export default function GenerateSeedPhrase() {
             <div className="onboarding-setup__page-with-header">
                 <div className="onboarding-setup__page-with-header__description">{t('recoveryPhrase.description')}</div>
                 <div className="p-10">
-                    <ControlledTextArea value={seedPhrase} readOnly />
+                    <ControlledTextArea
+                        value={seedPhrase}
+                        readOnly
+                        onCopy={(e) => {
```
orhoj, line 34: We agreed that it should possible to copy and paste the seed phrase into e.g. a password manager, so you should not block copying here.

shjortConcordium: We did? :thinking:

orhoj: Yes, for the above reason. Personally I think it would be fine to block it, but then people will probably just screenshot it instead, when they should just write it down by hand.

orhoj (approved):

PR #92, opened 2022-09-14
Fix incorrect cred number

shjortConcordium:
## Purpose

Fix incorrect cred number, due to providerIndex not being considered when finding next cred number.

## Changes

Moved two helper functions from recovery to helpers files.
bump version to 0.6.0 + update changelog.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/AddAccount/Confirm.tsx
+++ b/packages/browser-wallet/src/popup/pages/AddAccount/Confirm.tsx
@@ -15,6 +15,7 @@ import { absoluteRoutes } from '@popup/constants/routes';
 import { InternalMessageType } from '@concordium/browser-wallet-message-hub';
 import { popupMessageHandler } from '@popup/shared/message-handler';
 import { BackgroundResponseStatus } from '@shared/utils/types';
+import { isIdentityOfCredential } from '@shared/utils/identity-helpers';
 import { getGlobal, getNet } from '@shared/utils/network-helpers';
 import { addToastAtom } from '@popup/state';
 import AccountDetails from '../Account/AccountDetails';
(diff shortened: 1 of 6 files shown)


packages/browser-wallet/src/popup/pages/AddAccount/Confirm.tsx
```diff
@@ -57,7 +58,7 @@ export default function Confirm() {
             // Make request
             const expiry = Math.floor(Date.now() / 1000) + 720;
 
-            const credsOfCurrentIdentity = credentials.filter((cred) => cred.identityIndex === selectedIdentity.index);
+            const credsOfCurrentIdentity = credentials.filter(isIdentityOfCredential(selectedIdentity));
```
orhoj, line 61: So this line was the bug that Martin ran into and had Lars looking at in the mobile wallet?

shjortConcordium: Yes, one of them.

packages/browser-wallet/src/popup/shared/utils/account-helpers.ts
```diff
@@ -17,7 +17,7 @@ export function useIdentityOf(cred?: WalletCredential) {
         if (!cred) {
             return undefined;
         }
-        return identities.find(isIdentityOfCredential(id));
+        return identities.find((id) => isIdentityOfCredential(id)(cred));
```
orhoj, line 20: I was looking very hard at this previously, as I did not understand how `isIdentityOfCredential` would work here. Hehe...

shjortConcordium: Yeah :upside_down_face:

PR #94, opened 2022-09-14
Update recovery/account creation to align with mw

shjortConcordium:
## Purpose

Align some stuff regarding recovery and account creation with the mobile wallet.

## Changes

Increase maxEmpty gap to 20.
Use next available credNumber instead of highest + 1.
Check gaps in accounts during recovery.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/background/recovery.ts
+++ b/packages/browser-wallet/src/background/recovery.ts
@@ -11,10 +11,11 @@ import {
     Network,
 } from '@concordium/web-sdk';
 import { ExtensionMessageHandler, InternalMessageType } from '@concordium/browser-wallet-message-hub';
-import { BackgroundResponseStatus, IdentityIdentifier, RecoveryBackgroundResponse } from '@shared/utils/types';
+import { BackgroundResponseStatus, RecoveryBackgroundResponse } from '@shared/utils/types';
 import { Identity, CreationStatus, IdentityProvider, WalletCredential } from '@shared/storage/types';
 import { sessionIsRecovering, storedCredentials, storedCurrentNetwork, storedIdentities } from '@shared/storage/access';
 import { identityMatch, isIdentityOfCredential } from '@shared/utils/identity-helpers';
+import { getNextEmptyCredNumber } from '@popup/shared/utils/account-helpers';
 import { addCredential, addIdentity } from './update';
 import bgMessageHandler from './message-handler';
 import { openWindow } from './window-management';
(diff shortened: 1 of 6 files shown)


packages/browser-wallet/src/shared/utils/number-helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+    const sorted = used.sort((a, b) => a - b);
+    // If there are no credentials or 0 is not used, use 0
+    if (!sorted.length || sorted[0] > 0) {
+        return 0;
+    }
+    let i = 0;
+    while (i + 1 < sorted.length) {
+        // If the next credNumber is not the succesor of the current, there is a hole that can be filled
+        if (sorted[i] + 1 < sorted[i + 1]) {
+            break;
```
orhoj, line 14: With a for loop you could just return directly with the value here.

packages/browser-wallet/src/shared/utils/number-helpers.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Given a list of numbers return the smallest unused non-negative integer.
```
orhoj, line 2: It might also be a precondition that there are no duplicate values in the array.

orhoj (changes requested):

packages/browser-wallet/src/shared/utils/number-helpers.ts
shjortConcordium: Yeah, I guess it is, otherwise I would need to purge duplicates or adjust the condition. :thinking:

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -11,43 +11,47 @@ import {
...
-                providerIndex,
-            });
-        } else {
-            emptyIndices += 1;
+        // eslint-disable-next-line @typescript-eslint/no-loop-func
+        if (!existingCredentialsOfIdentity.some((cred) => cred.credNumber === credNumber)) {
+            const request = createCredentialV1({ ...credentialInput, credNumber });
+            const { credId } = request.cdi;
+            const accountInfo = await getAccountInfo(credId);
+            if (accountInfo) {
```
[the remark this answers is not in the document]
shjortConcordium: Yeah, we didn't do that before though :upside_down_face:

packages/browser-wallet/src/shared/utils/number-helpers.ts
shjortConcordium: Actually I don't? I tried testing it.

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -11,43 +11,47 @@ import {
...
 import { sessionIsRecovering, storedCredentials, storedCurrentNetwork, storedIdentities } from '@shared/storage/access';
 import { identityMatch, isIdentityOfCredential } from '@shared/utils/identity-helpers';
+import { getNextEmptyCredNumber } from '@popup/shared/utils/account-helpers';
 import { addCredential, addIdentity } from './update';
 import bgMessageHandler from './message-handler';
 import { openWindow } from './window-management';
 
 // How many empty identityIndices are allowed before stopping
-const maxEmpty = 10;
+const maxEmpty = 20;
```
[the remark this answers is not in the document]
orhoj: It is, but I asked him to set it to 10 when he made it 😄

Let's just stick with the 20, and then we can adjust if we need to in the future. I think we need a task for speeding up recovery in the browser though, as it's painfully slow compared to the mobile wallet.

packages/browser-wallet/src/background/recovery.ts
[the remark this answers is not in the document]
orhoj: And I refuse to fix mistakes of the past! 😄

packages/browser-wallet/src/shared/utils/number-helpers.ts
orhoj: Ah, sure, you would still need a return value outside of the loop in case the condition was never satisfied. I guess it's pretty much equal either way. Let's just keep it as is.

packages/browser-wallet/src/shared/utils/number-helpers.ts
orhoj: But you didn't add a test for it hehe. Looking at it again I think it works with duplicates too, not that we should have them for the current usage.

orhoj (approved):

PR #95, opened 2022-09-14
Use AES-GCM for encryption

orhoj:
## Purpose
Use AES-GCM for seed phrase encryption for proper checks of whether or not decryption went well. Uses browser libraries for decryption/encryption instead cryptojs.

NB: This is not compatible with existing seed phrases already in storage. Recover from a clean wallet.

## Changes
- Exchange encryption algorithm.
- Updated some components and atom usage as decryption and encryption is now `async`.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/shared/utils/account-helpers.ts
```diff
@@ -49,11 +49,11 @@ export function usePrivateKey(accountAddress: string | undefined): string | unde
...
-        if (!accountAddress || !credential || !identity) {
+        if (!accountAddress || !credential || !identity || seedPhrase.state !== 'hasData') {
             return undefined;
         }
 
-        return ConcordiumHdWallet.fromHex(seedPhrase, getNet(network))
+        return ConcordiumHdWallet.fromHex(seedPhrase.data, getNet(network))
             .getAccountSigningKey(identity.providerIndex, identity.index, credential.credNumber)
             .toString('hex');
     }, [credential?.credId, seedPhrase, identity?.index]);
```
shjortConcordium, line 59:
```suggestion
    }, [credential?.credId, seedPhrase.state, identity?.index]);
```
oder was?

orhoj: The value is present in the `data` subfield when `state === 'hasData'`.

shjortConcordium: Sure.
But don't we have problems with having objects in the dependency array?
And the assumption is that the data won't change?
Maybe this is fine actually

shjortConcordium (approved):

packages/browser-wallet/src/popup/shared/utils/account-helpers.ts
orhoj: I misread where the comment was placed. I will change it to state, and yes the assumption is that data won't change here.

PR #96, opened 2022-09-15
Change confirmation of identities/accounts

shjortConcordium:
## Purpose

Avoid an endless loop and instead check each individually when needed.

## Changes

On startup and when changing network, the background script starts listening to pending identities/accounts on the current network. When creating a new identity/account, start a "listener" to confirm it. 

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -0,0 +1,136 @@
...
+async function monitorIdentityStatus({ location, ...identity }: PendingIdentity, signal: AbortSignal) {
+    setTimeout(async function loop() {
+        if (signal.aborted) {
+            return;
+        }
+        let repeat = true;
+        try {
+            const response = (await (await fetch(location)).json()) as IdentityTokenContainer;
+            if (response.status === IdentityProviderIdentityStatus.Error) {
+                updateIdentities([
```
orhoj, line 79: Await the updates, otherwise you could end up setting repeat to false without having updated the identity.

packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -0,0 +1,136 @@
...
+                    },
+                ]);
+                repeat = false;
+            }
+        } finally {
+            if (repeat) {
+                setTimeout(loop, updateInterval);
+            }
+        }
+    }, 0);
```
orhoj, line 102: It feels hacky to set a 0 time timeout. Could you just define the `loop` function and then run it on the next line instead?

shjortConcordium: Yes.

PR #100, opened 2022-09-15
Fix recovery from settings crashing when there is an error

shjortConcordium:
## Purpose

Fix recovery from settings so that it fails gracefully, instead of exploding the entire wallet. (I also think the normal recovery would explode everything if that also failed)

## Changes

Ensure that we don't navigate from the RecoveryMain component before the seedPhrase loading atom is done loading.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
+++ b/packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState, useCallback } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { useAtom, useAtomValue } from 'jotai';
 import { useTranslation } from 'react-i18next';
-import { networkConfigurationAtom, seedPhraseAtom } from '@popup/store/settings';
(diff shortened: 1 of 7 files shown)


packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
```diff
@@ -25,21 +25,21 @@ export default function RecoveryMain() {
...
             navigate(absoluteRoutes.prompt.recovery.path, {
-                state: { status: BackgroundResponseStatus.Error, reason },
+                state: { payload: { status: BackgroundResponseStatus.Error, reason } },
             }),
         []
     );
 
     useEffect(() => {
-        if (runRecovery && !providers.length) {
+        if (runRecovery && !providers.length && seedPhrase.state === 'hasData') {
```
orhoj, line 34: I don't get why you need to wait for the seed phrase to be present to get the identity providers. They seem independent.

shjortConcordium: Yeah, that was to avoid crashing the wallet when this failed.

orhoj (approved):

PR #112, opened 2022-09-22
Fix undefined signal

shjortConcordium:
## Purpose

Ensure identityaccount creation doesnt not fail to confirm due to `undefined.signal`

## Changes

- Remove abortController.
- status monitors are now given the current network's genesisHash and terminate the loop if the current network has a different genesisHash.
- change update function to take the genesisHash as a parameter instead of loading it themselves. (To avoid loading the wrong network if it was changed while checking the status)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/background/index.ts
+++ b/packages/browser-wallet/src/background/index.ts
@@ -13,7 +13,7 @@ import {
 } from '@shared/storage/access';
 
 import JSONBig from 'json-bigint';
-import { ChromeStorageKey } from '@shared/storage/types';
+import { ChromeStorageKey, NetworkConfiguration } from '@shared/storage/types';
 import bgMessageHandler from './message-handler';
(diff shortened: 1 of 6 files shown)


packages/browser-wallet/src/background/index.ts
```diff
@@ -95,17 +95,21 @@ const injectScript: ExtensionMessageHandler = (_msg, sender, respond) => {
...
 
 chrome.storage.local.onChanged.addListener((changes) => {
     if (ChromeStorageKey.NetworkConfiguration in changes) {
-        networkChangeHandler();
+        networkChangeHandler(changes[ChromeStorageKey.NetworkConfiguration].newValue);
     }
 });
 
 chrome.runtime.onStartup.addListener(startupHandler);
-chrome.runtime.onInstalled.addListener(startupHandler);
```
orhoj, line 108: I think we need to keep this for updates.

orhoj (approved):

PR #115, opened 2022-09-22
Remove parameters from smart contract payload types in api

shjortConcordium:
## Purpose

Fix misleading types in the API, which lead users to believe they could supply the parameters for a smart contract as a buffer in the payload, when this is ignored, because we want them not serialized, so we can display them.

## Changes

- Change the typing
- update changelog
- bump api-helpers version

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api-helpers/CHANGELOG.md
+++ b/packages/browser-wallet-api-helpers/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.2.1
+
+### Changed
+
+-   Fixed broken link + typos in README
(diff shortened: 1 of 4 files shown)


packages/browser-wallet-api-helpers/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.2.1
+
+### Changed
+
+-   Fixed broken link + typos in README
+-   Removed parameters from smart contract types' payloads, due the wallet ignoring it in favor of separate arguments.
```
orhoj, line 8: This change will give type checking errors, so I think it should be a new major version.

orhoj (approved):

PR #118, opened 2022-09-26
Wrong selected identity when having failed identities

orhoj:
## Purpose
Fix an issue in account creation where another identity than the one chosen was shown.

## Changes
- Fix the identity index in `ChooseIdentity.tsx`. As the index is the index in the total list, and not just of confirmed identities, we have to ensure that we process all identities when building the list.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/AddAccount/ChooseIdentity.tsx
+++ b/packages/browser-wallet/src/popup/pages/AddAccount/ChooseIdentity.tsx
@@ -40,21 +40,24 @@ export default function ChooseIdentity() {
     return (
         <div className="flex-column align-center">
             <p className="m-t-20">{t('chooseIdentity')}</p>
-            {identities
-                .filter((identity) => identity.status === CreationStatus.Confirmed)
-                .map((identity, i) => (
-                    <IdCard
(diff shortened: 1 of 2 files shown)


packages/browser-wallet/src/popup/pages/AddAccount/ChooseIdentity.tsx
```diff
@@ -40,21 +40,24 @@ export default function ChooseIdentity() {
...
+                if (identity.status === CreationStatus.Confirmed) {
+                    return (
+                        <IdCard
+                            name={identity.name}
+                            key={`${identity.providerIndex}-${identity.index}`}
+                            provider={<IdentityProviderIcon provider={findProvider(identity)} />}
+                            status={identity.status}
+                            className="m-t-10"
+                            onClick={() => {
+                                setSelectedIdentityIndex(i);
```
soerenbf, line 53: I don't know if this makes sense, but would it have the same effect to use the index from `identity.index`? Would be better practice to use something that doesn't depend on the order or filters applied to the list of identities.

soerenbf (approved):

PR #119, opened 2022-09-26
Cleanup bg listeners

soerenbf:
## Purpose

Chrome documentation recommends adding event listeners synchronously, to ensure the service worker properly wakes up on events when in an inactive state. This PR ensure the event listeners used in identity issuance are properly added. Furthermore, hooks on the "chrome.webRequest" API do NOT currently wake up the service worker (a bug, which will be fixed in chrome v107: https://bugs.chromium.org/p/chromium/issues/detail?id=1024211#c73). A hack to solve this is also part of this PR.

## Changes

- Ensure event listeners are registered sync. at the top level of the script
- Added a hack to ensure service worker is kept alive for the duration of the corresponding ongoing identity issuance flow

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/background/identity-issuance.ts
```diff
@@ -12,85 +12,106 @@ import { confirmIdentity } from './confirmation';
...
+    // TODO this hack is only needed due to a bug in chrome, which prevents chrome.webRequest hooks from waking up the service worker from inactive state.
+    // Bug will be fixed in chrome v107
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=1024211#c73
+    chrome.alarms.clear(IDP_ALARM_NAME);
 
-async function startIdentityIssuance({
-    baseUrl,
-    ...identityRequestInputs
-}: IdentityRequestInput & { baseUrl: string }) {
     const network = await storedCurrentNetwork.get();
```
orhoj, line 23: Previously the network was loaded once and given as static input to the `respond` function. The issue with loading it dynamically each time is that the network could have changed, and you end up putting the identity into the wrong network.

packages/browser-wallet/src/background/identity-issuance.ts
```diff
@@ -12,85 +12,106 @@ import { confirmIdentity } from './confirmation';
...
+            sessionIdpTab.remove();
+        }
+
+        respond({
+            status: BackgroundResponseStatus.Success,
+            result: redirectUrl.substring(redirectUrl.indexOf(codeUriKey) + codeUriKey.length),
+        });
+    };
+
+    chrome.webRequest.onBeforeRequest.addListener(
```
orhoj, line 78: No more on redirect listener? Have you checked that we don't need that for Notabene and DTS?

orhoj:
I think that @soer8647 should also take a look at this.

packages/browser-wallet/src/background/identity-issuance.ts
```diff
@@ -12,85 +12,106 @@ import { confirmIdentity } from './confirmation';
...
+    await openWindow();
+    bgMessageHandler.sendInternalMessage(InternalMessageType.EndIdentityIssuance, { ...response, status });
+};
+
+export function addIdpListeners() {
+    const closedListener = async (tabId: number) => {
+        const idpTabId = await sessionIdpTab.get();
+
+        if (idpTabId !== undefined && tabId === idpTabId) {
+            chrome.tabs.onRemoved.removeListener(closedListener);
```
shjortConcordium, line 55: Now that this is a top-level listener, I don't think it should remove itself? Wouldn't this give an issue if I abort twice in a row?

shjortConcordium: Shouldn't this reset the idpTabId instead? (I guess it only matters if chrome would reuse tab ids)

packages/browser-wallet/src/background/identity-issuance.ts
```diff
@@ -12,85 +12,106 @@ import { confirmIdentity } from './confirmation';
...
+            });
         }
-        await openWindow();
-        bgMessageHandler.sendInternalMessage(InternalMessageType.EndIdentityIssuance, { ...response, status });
     };
 
+    chrome.tabs.onRemoved.addListener(closedListener);
+
+    const handleIdpRequest = (redirectUrl: string, tabId: number) => {
+        chrome.tabs.onRemoved.removeListener(closedListener);
```
shjortConcordium, line 65: Again, why are we removing this listener?

packages/browser-wallet/src/background/identity-issuance.ts
soerenbf: Yes, they all work. Redirect triggers onBeforeRequest as well.

packages/browser-wallet/src/background/identity-issuance.ts
orhoj: It might suffice, but I dislike that you have two values now that can be updated separately. I think the better short-term solution would be to have a single session value that contains the pending identity _and_ the network it is for. We really need to more sure that this is super robust and that we don't end up mixing network keys in the future.

What do you think @shjortConcordium?

shjortConcordium: Yeah, I don't see a reason not to mash the session values together.

shjortConcordium: I guess the only thing that makes me doubt the robustness is that you can keep the tab open while switching network, but I guess we should make sure to reset the tabId when switching network. If we don't already.

soerenbf: > I guess the only thing that makes me doubt the robustness is that you can keep the tab open while switching network, but I guess we should make sure to reset the tabId when switching network. If we don't already.

As I see it, this won't be an issue when we store the network with the pending identity, as that will also be the network used to store the identity.

soerenbf: Consolidated the pending identity props into one entry in sessions storage now.

soerenbf:
Implemented a more robust solution to keeping service worker alive for the duration of identity issuance flows.

orhoj:
LGTM.

shjortConcordium (approved):
LGTM and seems to work.

orhoj (approved):

PR #120, opened 2022-09-27
Alarm based entity monitoring

soerenbf:
## Purpose

Refactor monitoring of identities and credentials to be based on the `chrome.alarms` API, as recommended in the official documentation. This ensures that the service worker is reactivated if it is, for some reason, inactivated while still awaiting finalization of pending entities.

## Changes

- Change confirmation loops to `chrome.alarms` API

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -15,129 +16,134 @@ const isPendingCred = (cred: WalletCredential): cred is PendingWalletCredential
...
+const CRED_ALARM = 'check_cred';
+
+type ShouldRepeat = boolean;
+type CheckFun = () => Promise<ShouldRepeat>;
+
+const checkStatus = (checkFun: CheckFun, alarmId: string) => async () => {
+    const repeat = await checkFun();
+
+    if (!repeat) {
+        chrome.alarms.clear(alarmId);
```
shjortConcordium, line 31: If I make multiple identities, does this trigger multiple confirmations for every pending identity?
+ Couldn't there be a "race condition" where this clears the alarm, but between this checking for pending and getting here, I have created another identity, which confirmation is then cleared by this?

shjortConcordium: It seems to replace the alarm, and not duplicate it.

soerenbf: Yes, that's also what the documentation states:

https://developer.chrome.com/docs/extensions/reference/alarms/#method-create

shjortConcordium: Could the race condition not still occur? :thinking:

soerenbf: I suppose it does. I'll see if I can come up with a solution. Can probably use the alarm name to hold some sort of state for both identities and credentials.

shjortConcordium: Ideally I wouldn't have rewritten the confirmation loops to not be per identity/credential, and then you could do encode the identifier into the name, yes.
But with this I guess you could just not have a repeating alarm, but instead here schedule the next triggering, if there is still pending, That would remove that race condition, and worse case would be this delaying the next check.

soerenbf: > Ideally I wouldn't have rewritten the confirmation loops to not be per identity/credential, and then you could encode the identifier into the name, yes.

My brain hurts trying to understand the double negative...

I have to rethink the solution anyway and not use alarms, due to their unreliability for these kinds of jobs. Maybe we could go for a solution that periodically checks while the popup is open? In practice, I don't think it would make much of a difference anyway?

PR #123, opened 2022-09-27
Improve affordance of light/dark mode toggle

orhoj:
## Purpose
Improve affordance of light/dark mode toggle.

I think a slide that moves between the sun and the moon would be better, but it would take more work than this solution. Also toggle buttons seem to be tricky to get right: https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan?rq=1. 

## Changes
- Use new icons provided by Jens.
- Explicitly state the action of the button that is pressed, i.e. switching to dark/light mode.

## Checklist

      hard-to-understand areas.


soerenbf (changes requested):
The toggle doesn't really look like something you can interact with anymore, except for the cursor changing when hovering.

The icons make sense, however I think it would be nice if they were placed on top of an actual switch (or `ToggleCheckbox`). What do you think?

If you think it's a good idea, I can take a look at it, if you'd prefer that?

orhoj:
> The toggle doesn't really look like something you can interact with anymore, except for the cursor changing when hovering.
> 
> The icons make sense, however I think it would be nice if they were placed on top of an actual switch (or `ToggleCheckbox`). What do you think?
> 
> If you think it's a good idea, I can take a look at it, if you'd prefer that?

As I wrote in the description I think that would be nice, but this is just a quick-fix to go into the MVP. Now it is very explicit what it does, and before people were afraid of what the toggle did. If you can do something today, without it affecting the completion of the other bugs, then go for it.

soerenbf (approved):

soerenbf:
> > The toggle doesn't really look like something you can interact with anymore, except for the cursor changing when hovering.
> > The icons make sense, however I think it would be nice if they were placed on top of an actual switch (or `ToggleCheckbox`). What do you think?
> > If you think it's a good idea, I can take a look at it, if you'd prefer that?
> 
> As I wrote in the description I think that would be nice, but this is just a quick-fix to go into the MVP. Now it is very explicit what it does, and before people were afraid of what the toggle did. If you can do something today, without it affecting the completion of the other bugs, then go for it.

Added icons to the toggle. Please review and see if you agree on the solution 😄

PR #124, opened 2022-09-28
Persistent recovery

shjortConcordium:
## Purpose

Rework recovery, so it continues even after 5 minutes. (when the SW is killed)

## Changes

_Describe the changes that were needed.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/shared/storage/types.ts
```diff
@@ -158,3 +159,22 @@ export interface NetworkConfiguration {
     jsonRpcUrl: string;
     explorerUrl: string;
 }
+
+export type RecoveryPayload = {
+    providers: IdentityProvider[];
+    globalContext: CryptographicParameters;
+    seedAsHex: string;
```
orhoj, line 166: You already have the session password, so you should decrypt the seed phrase instead of saving it again here.

packages/browser-wallet/src/popup/store/identity.ts
```diff
@@ -42,3 +49,11 @@ export const identityNamesAtom = selectAtom(identitiesAtom, (identities) => {
 });
 
 export const isRecoveringAtom = atomWithChromeStorage<boolean>(ChromeStorageKey.IsRecovering, false, true);
+export const recoveryStatusAtom = atomWithChromeStorage<RecoveryStatus | undefined>(
+    ChromeStorageKey.RecoveryStatus,
+    undefined,
+    true
+);
+export const setRecoveryPayloadAtom = atom<null, RecoveryPayload, Promise<void>>(null, (_, set, payload) =>
+    set(recoveryStatusAtom, { payload })
```
orhoj, line 58: Does this only update the payload part of the status?

shjortConcordium: No, it resets everything else.

packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
```diff
@@ -51,15 +56,15 @@ export default function RecoveryMain() {
...
+                setRecoveryStatus({
                     providers,
                     globalContext: global,
                     seedAsHex: seedPhrase,
                     net: getNet(network),
-                });
-            })
+                })
+            )
+            .then(() => setIsRecovering(true))
```
orhoj, line 67: Could this be moved into the previous then as it was before? I don't understand why it was moved out.

packages/browser-wallet/src/background/index.ts
```diff
@@ -109,6 +109,13 @@ chrome.storage.local.onChanged.addListener((changes) => {
     }
 });
 
+chrome.storage.session.onChanged.addListener((changes) => {
+    if (ChromeStorageKey.IsRecovering in changes) {
```
orhoj, line 113: Should it check the value, or should it call `startRecovery` even when `IsRecovering === false`.

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -165,12 +220,34 @@ async function performRecovery({ providers, ...recoveryInputs }: Payload) {
...
-    const respond = async (result: RecoveryBackgroundResponse) => {
-        await openWindow();
-        bgMessageHandler.sendInternalMessage(InternalMessageType.RecoveryFinished, result);
-    };
-    performRecovery(msg.payload)
-        .then((added) => respond({ status: BackgroundResponseStatus.Success, added }))
-        .catch((e) => respond({ status: BackgroundResponseStatus.Error, reason: e.toString() }));
-};
+export function startRecovery() {
+    chrome.alarms.create(RECOVERY_ALARM_NAME, { delayInMinutes: 0 });
```
orhoj, line 224: From the documentation it says:
```
That is, setting delayInMinutes or periodInMinutes to less than 1 will not be honored and will cause a warning. 
```
so perhaps this should not be set to `0`.

packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
shjortConcordium: I changed it so that we wait setIsRecovering after setRecoveryStatus has been completed.
I can rewrite it to use async/await instead :nauseated_face:

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -165,12 +220,34 @@ async function performRecovery({ providers, ...recoveryInputs }: Payload) {
...
+        }
+        const isRecovering = await sessionIsRecovering.get();
+        if (isRecovering) {
+            // We use a lock to ensure only 1 recovery instance runs
+            navigator.locks.request(RECOVERY_LOCK, { ifAvailable: true }, (lock) => {
+                if (!lock) {
+                    // The lock was not granted - get out fast.
+                    return Promise.resolve();
+                }
+                chrome.alarms.create(RECOVERY_ALARM_NAME, { delayInMinutes: 5.1, periodInMinutes: 1 });
```
orhoj, line 240: Should `delayInMinutes` not be `< 5` minutes?

shjortConcordium: The way I understand it now is that the alarm can be used to revive the SW, not keep it alive. So I want it to happen right after the SW dies, not before.

packages/browser-wallet/src/background/recovery.ts
shjortConcordium: Yeah, I guess it only works properly now because it is unpacked, rewriting this.

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -4,36 +4,43 @@ import {
...
-import { addCredential, addIdentity } from './update';
+import { partition } from 'wallet-common-helpers';
+import { addCredential, addIdentity, updateCredentials } from './update';
 import bgMessageHandler from './message-handler';
 import { openWindow } from './window-management';
 
 // How many empty identityIndices are allowed before stopping
 const maxEmpty = 20;
+const RECOVERY_ALARM_NAME = 'recoveryAlarm';
+const RECOVERY_LOCK = 'recoveryLock';
```
orhoj, line 31: The current lock naming convention is:
```suggestion
const RECOVERY_LOCK = 'concordium_recovery_lock';
```
Perhaps all the lock constants should be collected to a single file (at some point in the future).

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -45,7 +52,7 @@ async function recoverAccounts(
                     credId,
                     credNumber,
                     status: CreationStatus.Confirmed,
-                    identityIndex,
+                    identityIndex: status.identityIndex || 0,
```
orhoj, line 55: I would define:
```
const identityIndex = status.identityIndex || 0;
```
to keep it consistent with what you do for `emptyIndices` and `credNumber` above.

packages/browser-wallet/src/background/recovery.ts
```diff
@@ -75,11 +81,18 @@ function getRecoverUrl(inputs: Omit<IdentityRecoveryRequestInput, 'timestamp' |
     return `${provider.metadata.recoveryStart}?${searchParams.toString()}`;
 }
 
-async function performRecovery({ providers, ...recoveryInputs }: Payload) {
+async function performRecovery() {
```
orhoj, line 84: Now is probably not the time, but this method should be split into multiple methods. It is quite difficult to get a good overview of what is happening inside.

packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
orhoj: But it should be a `then()` to the `setRecoveryStatus` then, right?

packages/browser-wallet/src/background/recovery.ts
shjortConcordium: Yeah, that seems like a dangerous refactoring to do **now** :thinking:

packages/browser-wallet/src/shared/utils/crypto.ts
```diff
@@ -11,11 +11,9 @@ const keyDerivationMethod = 'PBKDF2WithHmacSHA256';
 async function deriveKey(password: string, salt: Buffer): Promise<CryptoKey> {
     const passwordBuffer = Buffer.from(password, 'utf-8');
 
-    const baseKey = await global.crypto.subtle.importKey('raw', passwordBuffer, { name: 'PBKDF2' }, false, [
```
orhoj, line 14: Why are you removing `global` here? I actually thought that was required... :confused:

packages/browser-wallet/src/shared/storage/types.ts
```diff
@@ -17,6 +17,7 @@ export enum ChromeStorageKey {
     IsCreatingCredential = 'IsCreatingCredential',
     HasBeenOnboarded = 'hasBeenOnboarded',
     OnboardingLocation = 'onboardingLocation',
+    RecoveryStatus = 'RecoveryStatus',
```
orhoj, line 20: Naming convention is `camelCase`.
```suggestion
    RecoveryStatus = 'recoveryStatus',
```

packages/browser-wallet/src/popup/shared/utils/seedPhrase-helpers.ts
```diff
@@ -3,7 +3,7 @@ import { useAtomValue } from 'jotai';
 import { encryptedSeedPhraseAtom, sessionPasscodeAtom } from '@popup/store/settings';
 import { noOp, useAsyncMemo } from 'wallet-common-helpers';
 import { mnemonicToSeedSync } from '@scure/bip39';
-import { decrypt } from '../crypto';
+import { decrypt } from '@shared/utils/crypto';
```
orhoj, line 6: If you're fixing file placements, then you might as well also re-name this file to:
```
seed-phrase-helpers.ts
```

packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
```diff
@@ -51,24 +54,27 @@ export default function RecoveryMain() {
...
-    }, [runRecovery, isRecovering.loading, isRecovering.value, providers.length, seedPhrase]);
+    }, [runRecovery, isRecovering.loading, isRecovering.value, providers.length]);
 
     return (
         <>
             <PageHeader>{t('main.title')}</PageHeader>
             <div className="recovery__main onboarding-setup__page-with-header">
                 <div className="onboarding-setup__page-with-header__description">{t('main.description')}</div>
                 <PendingArrows className="identity-issuance__start__loading-arrows" />
+                <p>{recoveryStatus.value?.identityIndex}</p>
```
orhoj, line 74: Remember to clean up the status thing here.

packages/browser-wallet/src/popup/pages/Recovery/RecoveryMain.tsx
orhoj: I think I missed a parenthesis when reading it. But I like the changed version.

packages/browser-wallet/src/shared/utils/crypto.ts
shjortConcordium: I don't remember why but global is translated to window during building, and window is not defined in a service worker. :shrug: 
And it seems to work anyway :upside_down_face:

shjortConcordium: Generally I don't think you need to specify global unless to circumvent a shadowed variable :thinking:

orhoj: Ah, I guess that makes sense. It seems to work so it should be fine.

PR #125, opened 2022-09-28
Change wallet-api imports so it doesn't load wasm

shjortConcordium:
## Purpose

Fix so inject script doesn't cause errors when you use websites that dont allow wasm scripts.

## Changes

wallet-api now imports directly from common-sdk to avoid initializing the wasm module.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/package.json
```diff
@@ -34,6 +34,7 @@
         "react-virtualized-auto-sizer": "^1.0.6",
         "react-window": "^1.8.7",
         "react-window-infinite-loader": "^1.0.8",
+        "stream": "^0.0.2",
```
orhoj, line 37: Why is this needed now?

orhoj: I guess we don't get it indirectly from common-sdk or something.

shjortConcordium: Yeah we get it from common-sdk through the api now.

orhoj:
When I connect to a dApp (piggybank), then I get this error:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'call')
```

shjortConcordium:
> When I connect to a dApp (piggybank), then I get this error:
> 
> ```
> Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'call')
> ```

Turns out the "stream" package is **_very_** outdated.

orhoj (approved):

PR #126, opened 2022-09-29
Resume confirmation of entities when extension opens

soerenbf:
## Purpose

As we can never be sure that tasks created in the background script will finish, we need to ensure that they are resumed when it makes sense.

When it comes to confirmation of identities and credentials, the process will now resume, if it was unfinished, when the user opens the extension.

## Changes

- Resume confirmation loop when extension opens
- Keep track of which entities are already being monitored

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -7,53 +7,92 @@ import {
...
+    };
+    await run();
+};
+
+let monitoredCredentials: PendingWalletCredential[] = [];
+
+const isCredEqualTo = (c1: PendingWalletCredential) => (c2: PendingWalletCredential) => c1.credId === c2.credId;
+
+const isMonitoringCred = (id: PendingWalletCredential): boolean =>
+    monitoredCredentials.find(isCredEqualTo(id)) !== undefined;
```
shjortConcordium, line 51:
```suggestion
    monitoredCredentials.some(isCredEqualTo(id));
```

packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -62,57 +101,68 @@ async function monitorCredentialStatus(
...
+        ]);
     }
 }
 
+let monitoredIdentities: PendingIdentity[] = [];
+
+const isIdEqualTo = (id1: BaseIdentity) => (id2: BaseIdentity) =>
+    id1.index === id2.index && id1.providerIndex === id2.providerIndex;
+
+const isMonitoringIdentity = (id: PendingIdentity): boolean => monitoredIdentities.find(isIdEqualTo(id)) !== undefined;
```
shjortConcordium, line 117:
```suggestion
const isMonitoringIdentity = (id: PendingIdentity): boolean => monitoredIdentities.some(identityMatch(id));
```
We already have that helper somewhere called identityMatch.

packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -7,53 +7,92 @@ import {
...
-) {
+type ShouldLoop = boolean;
+
+/**
+ * Creates a promise, that resolves after given timeout
+ */
+const timer = (timeout: number): Promise<void> =>
+    new Promise((resolve) => {
+        setTimeout(resolve, timeout);
+    });
```
shjortConcordium, line 32:
```suggestion
import { sleep } from 'wallet-common-helpers';
```

packages/browser-wallet/src/background/index.ts
```diff
@@ -111,6 +111,10 @@ chrome.storage.local.onChanged.addListener((changes) => {
 
 chrome.runtime.onStartup.addListener(startupHandler);
 chrome.runtime.onInstalled.addListener(startupHandler);
+bgMessageHandler.handleMessage(createMessageTypeFilter(InternalMessageType.PopupReady), (_i, _s, respond) => {
```
shjortConcordium, line 114: So before the popup sent out the PopupReady to no one, and didn't expect an answer? (Why are we now telling it to wait for a response then?)

soerenbf: PopupReady was being used when spawning a new window to ensure the window is actually available and ready, before we try to trigger stuff in the popup from the background script (`background/window-management.ts` line 39).

But I guess you have a point that I shouldn't send it an answer, but merely start the confirmation process.

packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -7,53 +7,82 @@ import {
...
+    monitoredCredentials.push(cred);
+
+    const { deploymentHash, ...info } = cred;
+
+    await loop(async () => {
         const network = await storedCurrentNetwork.get();
         // Stop if the network has changed
         if (!network || network.genesisHash !== genesisHash) {
-            return;
+            monitoredCredentials = [];
```
shjortConcordium, line 57:
```suggestion
```
I don't think this works properly, actually. If I change network, and then this loops for the final time, it resets your list here, removing those added for the other network, right?

If you just remove this, it will resolve and then remove itself from the list afterwards, right?

shjortConcordium: Same for identities.

shjortConcordium: It is a bit hard to actually test without stagenet though :thinking:

soerenbf: haha, yes.. I had something else that would reset the lists when changing network  as part of the network change handler.. the current implementation doesn't make any sense, you're right

soerenbf: but I still think the reset needs to happen, due to the possibility of collission between identities/credentials across networks?

shjortConcordium (commented):
Otherwise LGTM

packages/browser-wallet/src/background/confirmation.ts
shjortConcordium: Then have your monitoredX index under the network hash.

shjortConcordium (approved):

PR #127, opened 2022-09-29
Update web sdk dependency to 2.0.0

shjortConcordium:
## Purpose

Use new version of web sdk


shjortConcordium:
> No usage of `getIdCred` for recovery speedup?

I figured that should be done separately (or just not before persistent recovery was merged)

soerenbf (approved):
LGTM

PR #134, opened 2022-10-06
Store pending transactions

soerenbf:
## Purpose

Store any transaction submitted from the browser wallet in local storage until it is finalized.

I tried doing a quick fingerprint of the transaction list, but I don't know if there is a better way to do this, that is fast enough to run every render? The reason this needs to be changed is to support switching between accounts that have the same transaction at the top of their transaction list.

## Changes

- Add store for pending transactions
- Update pending transactions on ccd drop, send ccd, send transaction
- Remove pending transactions when any of them match any finalized transaction in transaction list
- Alter transaction list reset, to support switching between accounts with equal latest transaction.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -21,6 +21,7 @@ import {
     sessionRecoveryStatus,
     sessionIdpTab,
 } from '@shared/storage/access';
+import { pendingTransactions } from '@shared/storage/transactions';
```
orhoj, line 24: I think this should be in `@shared/storage/access` along with the rest to keep it all together.

packages/browser-wallet/src/shared/storage/transactions.ts
```diff
@@ -0,0 +1,22 @@
+import { BrowserWalletAccountTransaction } from '@popup/shared/utils/transaction-history-types';
+import { makeStorageAccessor } from './access';
+import { ChromeStorageKey, StoredBrowserWalletTransaction } from './types';
+
+export const pendingTransactions = makeStorageAccessor<StoredBrowserWalletTransaction[]>(
+    'local',
```
orhoj, line 6: We might want to consider just putting this in `session` storage. If the user closes their browser, then the transaction will probably finalize before they open their wallet again?

packages/browser-wallet/src/shared/storage/transactions.ts
```diff
@@ -0,0 +1,22 @@
...
+);
+
+export const fromStoredTransaction = (t: StoredBrowserWalletTransaction): BrowserWalletAccountTransaction => ({
+    ...t,
+    time: BigInt(t.time),
+    amount: BigInt(t.amount),
+    cost: t.cost === undefined ? undefined : BigInt(t.cost),
+});
+
+export const toStoredTransaction = (t: BrowserWalletAccountTransaction): StoredBrowserWalletTransaction => ({
```
orhoj, line 17: When serializing big integers we usually use our own JSONHelper to stringify/parse.

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
```diff
@@ -226,9 +246,12 @@ export default function TransactionList({ onTransactionClick }: TransactionListP
         getTransactions(address, transactionResultLimit, 'descending', fromId)
             .then((transactionResult) => {
                 setHasNextPage(transactionResult.full);
+                removePendingTransactions(transactionResult.transactions);
```
orhoj, line 249: A pending transaction is never removed if you don't get the transaction externally. What could happen is that you send a transaction and close your browser / or just switch to a different account. Before returning to your wallet/account you receive a lot of transactions, more than `transactionResultLimit`, and then the pending transaction will still be visible at the top. The user then has to scroll through their transactions for the pending one to be removed.

I believe that the desktop wallet solves this by listening for the status of outgoing transactions, and does not rely on it coming back from the wallet proxy.

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
```diff
@@ -291,16 +312,15 @@ export default function TransactionList({ onTransactionClick }: TransactionListP
             );
         }
     } else {
-        // If the first transaction in the list changes, then we need to reload the entire
-        // list, otherwise the infinite list component bugs out displaying the transactions
-        // correctly.
-        const listKey = transactions[0].transactionHash ?? transactions[0].id;
+        // Quick and dirty fingerprint of transactions, to ensure transaction list resets properly.
+        const listKey = `${allTransactions.length}${allTransactions.reduce((acc, cur) => (acc + cur.id) % 73, 0)}`;
```
orhoj, line 316: I would think that the previous key prepended with the account address should work instead of this.

packages/browser-wallet/src/popup/store/transactions.ts
```diff
@@ -0,0 +1,69 @@
...
+>(
+    (get) => {
+        const selectedAccount = get(selectedAccountAtom);
+        if (selectedAccount === undefined) {
+            return [];
+        }
+
+        return get(pendingTransactionsFamily(selectedAccount));
+    },
+    (get, set, update) => {
```
orhoj, line 50: This `atom` is only use as an atom value, so is the setter needed?

orhoj (changes requested):

packages/browser-wallet/src/shared/storage/transactions.ts
soerenbf: Sure, no need to run unnecessary cleanup.

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
soerenbf: What do you mean with the previous key? the key from the previous render?

I was getting lists that didn't properly reset their sizes due to the key not resetting properly. As far as I remember, this was only really an issue when changing between accounts which both had the same latest transaction in their list, so a combination of the latest transaction and the currently selected account is not sufficient.

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
soerenbf: Yes, good catch, I'll revise.

packages/browser-wallet/src/shared/storage/transactions.ts
soerenbf: Makes sense, I'll adapt.

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
shjortConcordium: I think he meant accountAddress + (transactions[0].transactionHash ?? transactions[0].id)

shjortConcordium: But why would that not work, when switching between accounts with the same latest transcaction, the address would change, causing a rerender, right?

soerenbf: No, because the transactions are fetched as a result of the account changing, meaning the following would happen:

- account changes, re-render list with key `{address}{pendingTxHash}`.
- get new transactions, list key remains the same, due to pending transactions not resetting (which is the case with remote transactions)

soerenbf: But I guess i could ignore the pending transactions, and make the key based on the first remote transaction.

packages/browser-wallet/src/popup/shared/utils/transaction-helpers.ts
```diff
@@ -89,3 +90,42 @@ export function getTransactionTypeName(type: AccountTransactionType): string {
...
+    status: TransactionStatus.Pending,
+    time: BigInt(Math.round(Date.now() / 1000)),
+    id: 0,
+    cost,
+    transactionHash,
+    fromAddress,
+    toAddress,
+});
+
+export const pendingFromAccountTransaction = (
```
shjortConcordium, line 115: Could we have a verb in this? :thinking:
(This name is confusing)

shjortConcordium (commented):
I haven't tested it thoroughly, I'll do that monday.

packages/browser-wallet/src/shared/storage/access.ts
```diff
@@ -54,7 +54,7 @@ export type IndexedStorageAccessor<V> = {
  * @param area storeage area to store value in
  * @param key key used to store value
  */
-const makeStorageAccessor = <V>(area: chrome.storage.AreaName, key: ChromeStorageKey): StorageAccessor<V> => {
+export const makeStorageAccessor = <V>(area: chrome.storage.AreaName, key: ChromeStorageKey): StorageAccessor<V> => {
```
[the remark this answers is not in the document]
soerenbf: No idea..

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionList.tsx
shjortConcordium: I don't understand this still. Doesn't the pending transactions also change? (Why are they not resetting?)

shjortConcordium: Is it just the case where you switch between the sender and recipient? :thinking:

shjortConcordium: You changed it :thinking:

shjortConcordium (approved):

PR #137, opened 2022-10-11
Change: disable mainnet connection

DOBEN:
## Purpose

Disable mainnet connection on wCCD example

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/wCCD/src/Root.tsx
+++ b/examples/wCCD/src/Root.tsx
@@ -4,6 +4,7 @@ import React, { useEffect, useState, useMemo, useCallback } from 'react';
 import { detectConcordiumProvider } from '@concordium/browser-wallet-api-helpers';
 import WCCD from './wCCD';
 import { state, State } from './utils';
+import { TESTNET_GENESIS_BLOCK_HASH } from './constants';
 
 /**
  * Connect to wallet, setup application state context, and render children when the wallet API is ready for use.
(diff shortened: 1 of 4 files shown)


examples/wCCD/src/Root.tsx
```diff
@@ -17,10 +17,26 @@ export default function Root() {
...
+
     useEffect(() => {
         detectConcordiumProvider()
             .then((provider) => {
                 // Listen for relevant events from the wallet.
+                provider.on('chainChanged', (genesisBlock) => {
+                    // Check if the user is connected to testnet by checking if the genesisBlock is the testnet one.
+                    // Throw a warning and disconnect if wrong chain. We only want to
+                    // allow users to interact with our testnet smart contracts.
+                    if (genesisBlock !== '4221332d34e1694168c2a0c0b3fd0f273809612cb13d000d5c2e00e85f50f796') {
```
abizjak, line 33: Is it possible to put this in a named, documented constant so that it does not appear misteriously in two places?

abizjak (approved):
This looks reasonable to me, but @shjortConcordium and @orhoj have to be the final judge :)

shjortConcordium (approved):
LGTM, although it is annoying that you don't get a "network changed" or "account disconnected" when changing to mainnet while your mainnet account is not connected. :thinking:
(But I guess that is an issue with the wallet itself. Otherwise you would have to check what the network is each time the account changes, which might be okay if we had an endpoint for that, that didn't require the dApp to query the node :cry:)

shjortConcordium:
> > LGTM, although it is annoying that you don't get a "network changed" or "account disconnected" when changing to mainnet while your mainnet account is not connected. thinking (But I guess that is an issue with the wallet itself. Otherwise you would have to check what the network is each time the account changes, which might be okay if we had a direct endpoint for that, one that didn't require the dApp to query the node cry)
> 
> Is such an API something that should be added to the wallet? If so, where does one record such desires?

I have added an issue in Jira.

PR #140, opened 2022-10-14
Display cis2 tokens

soerenbf:
## Purpose

Display cis-2 tokens in tokens list

## Changes

- Make tokens page the default page under accounts
- Display fungibles and collectibles according to cis2 metadata spec

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/pages/Account/Tokens/Tokens.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    accountAddress={account.address}
+                    contractIndex={t.contractIndex}
+                    token={t}
+                />
+            ))}
+        </>
+    );
+}
+
+function Collectibles({ account }: ListProps) {
```
shjortConcordium, line 99: We should still have a way to tell the user if the collectible is owned or not?  (Maybe unowned has a more grey background or something? :thinking: )

packages/browser-wallet/src/popup/store/token.ts
```diff
@@ -0,0 +1,58 @@
...
+    const [accountAddress, contractIndex, tokenIndex] = parts;
+
+    return atom<Promise<bigint>>((get) => {
+        const client = get(jsonRpcClientAtom);
+        return getTokenBalance(client, contractIndex, tokenIndex, accountAddress);
+    });
+});
+
+export const tokenBalanceFamily = (accountAddress: string, contractIndex: string, tokenIndex: string) =>
+    tbf(`${accountAddress}.${contractIndex}.${tokenIndex}`);
```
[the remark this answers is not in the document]
shjortConcordium: (Because the balanceOf takes a list)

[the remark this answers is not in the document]
soerenbf: Yes, would be a nice optimization. I'll adapt!

packages/browser-wallet/src/popup/store/token.ts
```diff
@@ -0,0 +1,58 @@
...
+                        ...token,
+                        metadata: tokenMetadata.value[token.metadataLink],
+                    };
+                })
+            )
+        ),
+    };
+});
+
+const tbf = atomFamily<string, Atom<Promise<bigint>>>((tokenId: string) => {
```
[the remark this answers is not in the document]
soerenbf: tokenBalanceFamily 😛 it's not meant to be used externally, so I thought it was reasonable to make an abbreviated version

packages/browser-wallet/src/popup/pages/Account/Tokens/Tokens.tsx
shjortConcordium: If you can fit "(ont onwed)" in there, then do it :+1:

packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -122,3 +123,87 @@ export async function getTokenMetadata(tokenUrl: string): Promise<TokenMetadata>
...
+ */
+const deserializeBalanceAmounts = (value: string): bigint[] => {
+    const buf = Buffer.from(value, 'hex');
+    const n = buf.readUInt16LE(0);
+    let cursor = 2; // First 2 bytes hold number of token amounts included in response.
+    const amounts: bigint[] = [];
+
+    // eslint-disable-next-line no-plusplus
+    for (let i = 0; i < n; i++) {
+        const end = buf.slice(cursor).findIndex((b) => b < 2 ** 7) + 1; // Find the first byte with most significant bit not set, signaling the last byte in the leb128 slice.
```
shjortConcordium, line 159:
```suggestion
        const end = buf.subarray(cursor).findIndex((b) => b < 2 ** 7) + 1; // Find the first byte with most significant bit not set, signaling the last byte in the leb128 slice.
```
slice is deprecated :thinking:

packages/browser-wallet/src/popup/store/token.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        if (tokenIds.length === 0) {
+            return {};
+        }
+
+        return getContractBalances(client, contractIndex, tokenIds, accountAddress);
+    });
+});
+
+export const contractBalancesFamily = (accountAddress: string, contractIndex: string) =>
```
shjortConcordium, line 116: So the family acts as a cache, basically. I guess we will need a way to invalidate the cache, because currently I can only force a reload of my assets by leaving the account page and coming back.
(But I assume we want it to be updated after some time :thinking:)

soerenbf: Yes, thought that would be fine for now to just update it when you navigate to the account page. An atom family is just a map of string to atom, so yes it basically acts like a cache. I don't know how it would make sense to update the values, as we don't have a good way of knowing if the balance has changed, except for doing contract invoke every x seconds, which is probably not a good idea since it uses NRG?

An alternative could be to update all cis-2 balances when the public CCD balance changes. Of course that would only update the balance if you're the sender of the cis-2 transfer.

shjortConcordium: I'm not sure I understand your concern, that it uses NRG? (it doesn't cost anything for the user)

It would be great if we could "invalidate the cache" after some time (like 2 minutes). Maybe that can be done in a separate PR.

soerenbf: Decided on just going with a 1 minute interval for updating, and scrapping the ccd balance effect.

shjortConcordium (approved):

PR #143, opened 2022-10-25
Pending transactions indexed by network

soerenbf:
## Purpose

Discovered some issues when setting up a new account on a freshly added network:

- pending transactions were not network specific
- pending transaction amount displayed wrong
- network specific atoms didn't update when changing network

## Changes

- Refactor pending transactions atom to set value through atom when possible
- Refresh atoms on network change
- Make `TransactionListElement` render pending transactions properly in terms of amount and cost both for sending and receiving accounts

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionElement.tsx
+++ b/packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionElement.tsx
@@ -177,6 +177,10 @@ export default function TransactionElement({ accountAddress, transaction, style,
     const transactionTime = withDate ? withDateAndTime(transactionDate) : onlyTime(transactionDate);
     const failed = transaction.status === TransactionStatus.Failed;
     const isSender = transaction.fromAddress === accountAddress;
+
+    // Flip the amount is selected account is sender, and amount is positive. We expect the transaction list endpoint to sign the amount based on this,
+    // but this is not the case for pending transactions. This seeks to emulate the behaviour of the transaction list endpoint.
+    // TODO: check that this still works when shield, unshield, and encrypted transfers are implemented.
     const amount = isSender && transaction.amount > 0n ? -transaction.amount : transaction.amount;
 
     return (
--- a/packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionElement.tsx
+++ b/packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionElement.tsx
@@ -176,6 +176,8 @@ export default function TransactionElement({ accountAddress, transaction, style,
     const transactionDate = dateFromTimestamp(transaction.time, TimeStampUnit.seconds);
     const transactionTime = withDate ? withDateAndTime(transactionDate) : onlyTime(transactionDate);
     const failed = transaction.status === TransactionStatus.Failed;
+    const isSender = transaction.fromAddress === accountAddress;
+    const amount = isSender && transaction.amount > 0n ? -transaction.amount : transaction.amount;
 
     return (
         <div
@@ -200,7 +202,7 @@ export default function TransactionElement({ accountAddress, transaction, style,
                             isGreenAmount(transaction, accountAddress) && 'transaction-element__amount__greenText'
                         )}
                     >
-                        {displayAsCcd(transaction.cost ? transaction.amount - transaction.cost : transaction.amount)}
+                        {displayAsCcd(transaction.cost && isSender ? amount - transaction.cost : amount)}
                     </p>
                 }
             />
@@ -212,7 +214,9 @@ export default function TransactionElement({ accountAddress, transaction, style,
                     </>
                 }
                 right={
-                    transaction.cost !== undefined ? buildFeeString(transaction.cost, accountAddress, transaction) : ''
+                    transaction.cost !== undefined && isSender
+                        ? buildFeeString(transaction.cost, accountAddress, { ...transaction, amount })
+                        : ''
                 }
             />
         </div>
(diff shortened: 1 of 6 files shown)


packages/browser-wallet/src/popup/store/utils.ts
```diff
@@ -91,7 +91,7 @@ export function atomWithChromeStorage<V>(key: ChromeStorageKey, fallback: V, wit
 
         const listener = (changes: Record<string, chrome.storage.StorageChange>) =>
             getStoredValue().then((value) => {
-                if (key in changes) {
+                if (key in changes || ChromeStorageKey.NetworkConfiguration in changes) {
```
shjortConcordium, line 94: Wait... Why do we getStoredValue and then inside check if the changes has anything to do with the current key?

shjortConcordium: (Instead of the opposite)

soerenbf: Yes.. good point...

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionElement.tsx
```diff
@@ -176,6 +176,8 @@ export default function TransactionElement({ accountAddress, transaction, style,
     const transactionDate = dateFromTimestamp(transaction.time, TimeStampUnit.seconds);
     const transactionTime = withDate ? withDateAndTime(transactionDate) : onlyTime(transactionDate);
     const failed = transaction.status === TransactionStatus.Failed;
+    const isSender = transaction.fromAddress === accountAddress;
+    const amount = isSender && transaction.amount > 0n ? -transaction.amount : transaction.amount;
```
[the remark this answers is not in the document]
soerenbf: Yes, I'm flipping the transaction amount for transactions with a positive `amount` when `fromAddress` matches the currently selected account. As far as I understand, this will never have any effect for transactions returned from the wallet proxy.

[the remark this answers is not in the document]
shjortConcordium: Wouldn't it make more sense to change the pending transaction's, when they are saved, to ensure that we don't have to do this flip?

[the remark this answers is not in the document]
shjortConcordium: It seems like it does work currently. I am just averse to this special handling, that relies on an assumption that the other cases (transactions from the wallet proxy) have a certain value? :thinking:

packages/browser-wallet/src/popup/store/transactions.ts
```diff
@@ -43,28 +44,51 @@ const monitorTransactionStatus = (genesisHash: string) => {
...
+                    const spt = useIndexedStorage(sessionPendingTransactions, async () => currentNetwork.genesisHash);
+                    spt.set(nextPending.map(stringify));
+                }
+            });
         }
     );
+
+    derived.onMount = (startMonitoring) => {
+        startMonitoring([]);
+    };
```
shjortConcordium, line 89: What does this do, if anything?

packages/browser-wallet/src/popup/store/transactions.ts
```diff
@@ -43,28 +44,51 @@ const monitorTransactionStatus = (genesisHash: string) => {
...
-        (_, set, update) => {
-            set(base, update.map(stringify));
+                const networkChanged = network.genesisHash !== currentNetwork.genesisHash;
+                const nextPending = pending.filter((p) => p.transactionHash !== transactionHash);
+
+                if (!networkChanged) {
+                    set(parsed, nextPending);
+                } else {
+                    const spt = useIndexedStorage(sessionPendingTransactions, async () => currentNetwork.genesisHash);
+                    spt.set(nextPending.map(stringify));
```
shjortConcordium, line 81: It seems weird that you do theses updates inside the forEach, instead of filtering and then doing a single update?

packages/browser-wallet/src/popup/store/transactions.ts
soerenbf: Well... it starts monitoring? 😄 

When the atom mounts (i.e. is used in a view), we check all the pending transactions stored in session. Imagine the use case where you

1. Submit a transaction
2. Close the wallet
3. Re-open the wallet

Here, we have to re-initiate the monitoring, as it stops when the user closes the wallet.

packages/browser-wallet/src/popup/store/transactions.ts
soerenbf: I've done it this way because I want the transactions to be removed as soon as they've been verified as finalized. As I read it, what you propose is to wait for all pending transactions to be finalized, and then remove them in one single update? Or am I misunderstanding?

packages/browser-wallet/src/popup/pages/Account/TransactionLog/TransactionElement.tsx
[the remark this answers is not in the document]
soerenbf: When transactions are fetched from the wallet proxy, the sign (+/-) is set based on whether the address used for querying the transactions is sender/receiver. What I'm doing is emulating this behaviour, as users can have both the sending/receiving account of a pending transaction.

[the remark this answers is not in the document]
shjortConcordium: Yes. Maybe add a TODO that reminds us to check it? :upside_down_face:

packages/browser-wallet/src/popup/store/transactions.ts
soerenbf: > The monitor() blocks until it is finalized?

Correct.

> I guess the only issue is that pending is not updated, so the updates will cancel each other out?

I don't understand? Why is pending not updated? That's what happens in the 5 lines you marked?

packages/browser-wallet/src/popup/store/transactions.ts
soerenbf: Added a comment that explains the behaviour

shjortConcordium (approved):

PR #147, opened 2022-10-31
Add voting example

### LINKED ISSUE #146: Add voting example from hackathon
**Task description**
Clean-up and add the voting example from the hackathon

DOBEN:
## Purpose

closes #146 

Adding voting project from the hackathon. Thanks to @Bargsteen @bisgardo @tschudid Kåre for the amazing team that created this demo.

Smart contract part can be found [here](https://github.com/Concordium/concordium-rust-smart-contracts/pull/189)

## Checklist

      hard-to-understand areas.


examples/voting/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:16-slim AS build
+WORKDIR /build
+COPY package.json package-lock.json ./
+RUN npm install
+COPY . .
+RUN npm run build
```
shjortConcordium, line 6: Is there a reason we use npm here, instead of yarn like we do elsewhere.
(I guess maybe it was because of the create-react-app stuff)
(I haven't tested that this works after my changes :upside_down_face: )

But I would assume we could use basically the same dockerfile as wccd's here. :thinking:

examples/voting/httpd.conf
```diff
@@ -0,0 +1,550 @@
+#
+# This is the main Apache HTTP server configuration file.  It contains the
```
shjortConcordium, line 2: And, do we need this? :thinking:

shjortConcordium (changes requested):
Assuming you agree with my changes, its basically just the dockerfile that I have anything for.

DOBEN:
> Assuming you agree with my changes, its basically just the dockerfile that I have anything for.

I love your changes :+1:

shjortConcordium:
I noticed that you could not access child pages in the example (like http://localhost:8080/results/1941), which I figured was because of I changed away from the create-react-app stuff. So I decided to fix that. And changed the package.json so we can avoid having a different one for the docker.

examples/voting/src/Wallet.js
```diff
@@ -0,0 +1,144 @@
...
+
+export async function connect(client, setConnectedAccount) {
+    const account = await client.connect();
+    return setConnectedAccount(account);
+}
+
+// Check if the user is connected to the testnet chain by checking if the testnet genesisBlock exists.
+// The smart contract voting module is deployed on the testnet chain.
+async function checkConnectedToTestnet(client) {
+    return await client
```
shjortConcordium, line 42:
```suggestion
/* eslint-disable consistent-return */
/* eslint-disable react/prop-types */
/* eslint-disable react/jsx-filename-extension */
/* eslint-disable import/no-unresolved */
import React from 'react';
import { detectConcordiumProvider } from '@concordium/browser-wallet-api-helpers';
import { Alert, Button } from 'react-bootstrap';
import { AccountTransactionType, GtuAmount, ModuleReference } from '@concordium/web-sdk';
import moment from 'moment';
import { RAW_SCHEMA_BASE64, TESTNET_GENESIS_BLOCK_HASH } from './config';

export async function init(setConnectedAccount) {
    const client = await detectConcordiumProvider();
    // Listen for relevant events from the wallet.
    client.on('accountChanged', (account) => {
        console.debug('browserwallet event: accountChange', { account });
        setConnectedAccount(account);
    });
    client.on('accountDisconnected', () => {
        console.debug('browserwallet event: accountDisconnected');
        client.getMostRecentlySelectedAccount().then(setConnectedAccount);
    });
    client.on('chainChanged', (chain) => {
        console.debug('browserwallet event: chainChanged', { chain });
    });
    client.getMostRecentlySelectedAccount().then(setConnectedAccount);

    return client;
}

export async function connect(client, setConnectedAccount) {
    const account = await client.connect();
    return setConnectedAccount(account);
}

// Check if the user is connected to the testnet chain by checking if the testnet genesisBlock exists.
// The smart contract voting module is deployed on the testnet chain.
function checkConnectedToTestnet(client) {
    return client
```

Removes the rule /* eslint-disable @typescript-eslint/return-await */
by removing the redundant await in the `return await client`

shjortConcordium (approved):

DOBEN:
@shjortConcordium @soerenbf I updated the front-end to work with the merged example voting smart contracts which have more basic getter functions compared to the original version of the smart contracts (the votes are not accumulated in the smart contract by one getter function anymore instead the votes can be retrieved for each voting option by calling a function in the smart contract.).

PR #157, opened 2022-11-09
Bugfix/update tokens

soerenbf:
## Purpose

Various fixes for the manage tokens flow:

- Contract select validation aligned and improved
- Show confirmation message in toast when updating tokens
- Show list status message when list is empty (either due to no results for search or missing metadata)
- Navigate to most relevant page, based on changes made to the list.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -77,26 +77,32 @@ function deserializeTokenMetadataReturnValue(returnValue: string): string[] {
...
-): Promise<string | undefined> {
+): Promise<CIS2ConfirmationError | undefined> {
     const supports = await client.invokeContract({
         contract: { index, subindex },
         method: `${contractName}.supports`,
         parameter: getCIS2Identifier(),
     });
     if (!supports || supports.tag === 'failure') {
-        return 'Chosen contract does not support CIS-0';
+        return CIS2ConfirmationError.Cis0Error;
```
shjortConcordium, line 98: Why not just use the i18n directly instead of creating this enum?

packages/browser-wallet/src/popup/shared/utils/validation-helpers.ts
```diff
@@ -0,0 +1,17 @@
+import { Validate, ValidateResult } from 'react-hook-form';
+import debounce from 'lodash.debounce';
+
+export type AsyncValidate<V> = (fieldValue: V) => Promise<ValidateResult>;
+
+export function debouncedAsyncValidate<V>(validator: AsyncValidate<V>, ms: number, leading = false): Validate<V> {
+    return async (value) =>
```
shjortConcordium, line 7:
```suggestion
    return (value) =>
```
?

packages/browser-wallet/src/popup/shared/utils/validation-helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+        new Promise((resolve) => {
+            debounce(
+                async (v: V) => {
```
shjortConcordium, line 10:
```suggestion
                (v: V) => {
```
?

shjortConcordium (approved):

packages/browser-wallet/src/shared/utils/token-helpers.ts
soerenbf: Because this is a "helper" file without access to translations. Unless theres a way to do that, which I don't know of?

PR #164, opened 2022-11-10
Reject prompt requests, if there is already a prompt open

shjortConcordium:
## Purpose

Ensure that we don't open multiple prompts at the same time.

## Changes

- Add variable to check if a prompt is already open, reject if it is sat.
- Reworked the type sent between the script.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/background/index.ts
+++ b/packages/browser-wallet/src/background/index.ts
@@ -3,6 +3,7 @@ import {
     InternalMessageType,
     MessageType,
     ExtensionMessageHandler,
+    MessageStatusWrapper,
 } from '@concordium/browser-wallet-message-hub';
 import { HttpProvider } from '@concordium/web-sdk';
 import {
@@ -10,13 +11,21 @@ import {
     storedSelectedAccount,
     storedCurrentNetwork,
     sessionPasscode,
+    sessionOpenPrompt,
 } from '@shared/storage/access';
 
 import JSONBig from 'json-bigint';
 import { ChromeStorageKey, NetworkConfiguration } from '@shared/storage/types';
 import { buildURLwithSearchParameters } from '@shared/utils/url-helpers';
 import bgMessageHandler from './message-handler';
-import { forwardToPopup, HandleMessage, HandleResponse, RunCondition, setPopupSize } from './window-management';
+import {
+    forwardToPopup,
+    HandleMessage,
+    HandleResponse,
+    RunCondition,
+    runConditionComposer,
+    setPopupSize,
+} from './window-management';
 import { addIdpListeners, identityIssuanceHandler } from './identity-issuance';
 import { startMonitoringPendingStatus } from './confirmation';
 import { sendCredentialHandler } from './credential-deployment';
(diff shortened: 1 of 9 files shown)


packages/browser-wallet/src/background/recovery.ts
```diff
@@ -36,7 +36,7 @@ import bgMessageHandler from './message-handler';
 import { openWindow } from './window-management';
 
 // How many empty identityIndices are allowed before stopping
-const maxEmpty = 20;
+const maxEmpty = 1;
```
soerenbf, line 39: Is this a test change that needs to be reverted?

packages/browser-wallet/src/background/window-management.ts
```diff
@@ -160,3 +166,16 @@ export const forwardToPopup = <P, R>(
...
+    firstHandler: RunCondition<R>,
+    secondHandler: RunCondition<R>
+): RunCondition<R> {
+    return async (msg, sender) => {
+        const first = await firstHandler(msg, sender);
+        if (!first.run) {
+            return first;
+        }
+        return secondHandler(msg, sender);
+    };
```
soerenbf, line 180: Couldn't we improve this to take an arbitrary number of `RunCondition`s?

packages/browser-wallet/src/background/index.ts
```diff
@@ -303,13 +314,48 @@ bgMessageHandler.handleMessage(
...
+    if (isPromptOpen) {
+        return { run: false, response: { success: false, message: 'Another prompt is already open' } };
+    }
+    sessionOpenPrompt.set(true);
+    return { run: true };
+};
+
+function withPromptEnd() {
+    sessionOpenPrompt.set(false);
+}
```
soerenbf, line 328: I could't do the following, but what is there any way this doesn't get triggered? Then the user will be in a state where no new requests can be opened.

Alternatively, we have a `background/window-management.ts@testPopupOpen`, which we can use, that doesn't depend on setting stuff on session storage, which could potentially block users.

shjortConcordium: I don't believe this can not be triggered (without closing the browser).

It is an interesting idea, but it would also block if you any window open. (the popup itself, or recovery results, etc.)
I'm don't know whether we want to do that.

packages/browser-wallet/src/background/window-management.ts
shjortConcordium: We could, but why?

soerenbf: Well... more reusable with no overhead? What's not to like? Anyway if you don't want to, idc 😉

packages/browser-wallet/src/background/index.ts
soerenbf: Yes, you're right.. don't know what I was thinking. Maybe instead of doing sessionstorage, could we do something similar to `testPoopupOpen` where we ping the prompt page to test? I think it's the safer approach of the two.

packages/browser-wallet/src/background/window-management.ts
shjortConcordium: I dont want to do this when we dont need it.

soerenbf (approved):

PR #165, opened 2022-11-10
Textual changes

shjortConcordium:
## Purpose

Update text based on feedback.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/AddAccount/i18n/en.ts
+++ b/packages/browser-wallet/src/popup/pages/AddAccount/i18n/en.ts
@@ -1,5 +1,5 @@
 const t = {
-    chooseIdentity: 'Which identity should create the account?',
+    chooseIdentity: 'Which identity do you want to use to create the account?',
     createAccount: 'Create new account',
     noIdentities: 'You have no identities, however you need an identity to create an account.',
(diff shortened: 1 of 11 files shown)


soerenbf (approved):

packages/browser-wallet/src/popup/pages/AddAccount/i18n/en.ts
```diff
@@ -1,5 +1,5 @@
 const t = {
-    chooseIdentity: 'Which identity should create the account?',
+    chooseIdentity: 'Which identity do you want to use to create the account?',
```
soerenbf, line 2: We need some padding on the element, where this is used.

soerenbf: And text center as well.

soerenbf (changes requested):

packages/browser-wallet/src/popup/pages/AddAccount/i18n/en.ts
shjortConcordium: nice catch!

soerenbf (approved):

PR #166, opened 2022-11-10
fall back to a sensible route on account change

soerenbf:
## Changes

On account change:
- Route to tokens root for token details
- Route to transaction list for transaction details
- Route to tokens for manage tokens

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/pages/Account/Tokens/Tokens.tsx
```diff
@@ -138,7 +138,6 @@ function Tokens() {
 export default function TokensRoutes() {
     const account = useSelectedCredential();
     const nav = useNavigate();
-    useAtomValue(tokensAtom); // Ensure tokens are kept in memory
```
shjortConcordium, line 141: This is not needed after all?

shjortConcordium (approved):

packages/browser-wallet/src/popup/pages/Account/Tokens/Tokens.tsx
soerenbf: Nope.. I had misunderstood something about how atoms work. Thought they were removed from memory when they unmount, but that doesn't seem to be the case.

PR #171, opened 2022-11-14
Improve handling of bad token transfer

shjortConcordium:
## Purpose

Block token transfers if simulation fails + improve error message.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Account/SendCcd/CreateTransfer.tsx
+++ b/packages/browser-wallet/src/popup/pages/Account/SendCcd/CreateTransfer.tsx
@@ -118,9 +118,9 @@ function CreateTransaction({ exchangeRate, tokens, setCost, setDetailsExpanded }
                     );
                     form.setValue('executionEnergy', energy.execution.toString());
                     return energy.total;
-                } catch {
-                    addToast(t('sendCcd.transferInvokeFailed'));
-                    return undefined;
+                } catch (e) {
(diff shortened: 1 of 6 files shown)


packages/browser-wallet/src/popup/pages/Account/SendCcd/CreateTransfer.tsx
```diff
@@ -118,9 +118,9 @@ function CreateTransaction({ exchangeRate, tokens, setCost, setDetailsExpanded }
                     );
                     form.setValue('executionEnergy', energy.execution.toString());
                     return energy.total;
-                } catch {
-                    addToast(t('sendCcd.transferInvokeFailed'));
-                    return undefined;
+                } catch (e) {
+                    addToast(t('sendCcd.transferInvokeFailed', { message: (e as Error).message }));
+                    return false;
```
soerenbf, line 123: This feels really dirty.. You use `undefined` to indicate malformed account address, false to indicate invoke failure, and bigint to indicate sucess. Combined with "falsy" comparisons, which in this case means `0n | undefined | false`, it gets a little hard to keep track of side effects.

Would be nice with something a little more explicit.

soerenbf (approved):

packages/browser-wallet/src/popup/pages/Account/SendCcd/CreateTransfer.tsx
shjortConcordium: Yeah, I added an explicit type, please check :sweat_smile:

PR #172, opened 2022-11-14
Improve handling of bad balanceOf on tokens

shjortConcordium:
## Purpose

- Filter tokens with failing `balanceOf` in Manage page 
- Show error when `balanceOf` fails 
- Show no balance instead of nothing in token list/detail if `balanceOf` fails.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/shared/utils/token-helpers.ts
+++ b/packages/browser-wallet/src/shared/utils/token-helpers.ts
@@ -386,14 +386,21 @@ export async function getTokenTransferEnergy(
     return { execution, total };
 }
 
-export const getTokens = async (
+type GetTokensResult = {
+    id: string;
+    metadataLink: string;
+    metadata: TokenMetadata | undefined;
+    balance: bigint;
+}[];
+
+export async function getTokens(
     contractDetails: ContractDetails,
     client: JsonRpcClient,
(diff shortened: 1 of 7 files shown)


packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -399,16 +402,25 @@ export const getTokens = async (
...
-    return ids.map((id, i) => ({
-        id,
-        metadataLink: metadataUrls[i],
-        metadata: metadata[i],
-        balance: balances[id] ?? 0n,
-    }));
+    return ids
+        .filter((id, i) => id in balances && metadata[i])
+        .map((id, i) => ({
+            id,
```
shjortConcordium, line 419: I realise that this needs to be rewritten,so the indices do not get shifted. (Probably gonna merge into a reduce)

soerenbf (approved):
LGTM, when you fix the `.filter.map` -> `.reduce`

packages/browser-wallet/src/shared/utils/token-helpers.ts
shjortConcordium: Is it okay @soerenbf?

PR #173, opened 2022-11-14
improve handling of bad tokenMetadata

shjortConcordium:
## Purpose

Add proper error when tokenMetadata fails during manage.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/shared/utils/token-helpers.ts
+++ b/packages/browser-wallet/src/shared/utils/token-helpers.ts
@@ -402,10 +402,22 @@ export async function getTokens(
     onError?: (error: string) => void
 ): Promise<GetTokensResult> {
     const metadataPromise: Promise<[string[], Array<TokenMetadata | undefined>]> = (async () => {
-        const metadataUrls = await getTokenUrl(client, ids, contractDetails);
-        const metadata = await Promise.all(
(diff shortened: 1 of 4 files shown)


packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -402,10 +402,22 @@ export async function getTokens(
...
+        }
+        let metadata;
+        try {
+            metadata = await Promise.all(
+                metadataUrls.map((url) => getTokenMetadata(url, network).catch(() => Promise.resolve(undefined)))
+            );
+        } catch (e) {
+            onError?.(`Failed to load a metadata file on index: ${contractDetails.index}`);
+            return [[], []];
+        }
```
soerenbf, line 420: There's already a catch on the `getTokenMetadata` request, so this should be redundant (and also a wrong implementation, when the expected outcome is to filter away individual tokens without metadata)?

soerenbf (changes requested):

packages/browser-wallet/src/shared/utils/token-helpers.ts
shjortConcordium: Yes, not sure what I was thinking. Reverted it.

soerenbf (approved):

PR #177, opened 2022-11-15
Fix token search, so the wallet doesn't crash

shjortConcordium:
## Purpose

Fix token search, so the wallet doesn't crash

## Changes

Fix token search, so it does not put undefined in the list, if there is no token.
 + fix the incorrect type.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
+++ b/packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
@@ -103,7 +103,7 @@ const lookupTokenIdConfigure = (
             ac = new AbortController();
             const { signal } = ac;
 
-            let value: ContractTokenDetails[] | undefined;
+            let value: ContractTokenDetails[] = [];
 
             if (!searchQuery) {
                 return;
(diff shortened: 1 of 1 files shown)


packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
```diff
@@ -113,8 +113,9 @@ const lookupTokenIdConfigure = (
...
             try {
                 const [token] = await getTokens(contractDetails, client, network, account, [searchQuery]);
-
-                value = [token as ContractTokenDetails];
+                if (token && token.metadata) {
+                    value = [token as ContractTokenDetails];
+                }
             } catch {
                 value = [];
             }
```
soerenbf, line 121: Seems a bit weird to re-set the `value` to the value it already has? Maybe change to finally, and move the section below up in that block instead?

soerenbf (approved):

packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
shjortConcordium: I think it is okay to be explicit here. :shrug:

PR #192, opened 2022-11-21
Fix incorrect id on search/searching on non-hex values

shjortConcordium:
## Purpose

Fix incorrect Id appearing on tokens when searching.

## Changes

Perform more check before searching for tokens

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
+++ b/packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
@@ -1,8 +1,8 @@
 import React, { CSSProperties, forwardRef, ReactNode, useCallback, useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
-import { isDefined, noOp, useUpdateEffect } from 'wallet-common-helpers';
-import { isHex, JsonRpcClient } from '@concordium/web-sdk';
+import { isHex, isDefined, noOp, useUpdateEffect } from 'wallet-common-helpers';
+import { JsonRpcClient } from '@concordium/web-sdk';
 import { useAtom, useAtomValue, useSetAtom } from 'jotai';
 import debounce from 'lodash.debounce';
 import InfiniteLoader from 'react-window-infinite-loader';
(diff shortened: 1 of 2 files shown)


packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
```diff
@@ -84,6 +84,10 @@ const InfiniteTokenList = forwardRef<HTMLDivElement, InfiniteTokenListProps>(
     }
 );
 
+function isEven(i: number) {
+    return i % 2 === 0;
+}
```
soerenbf, line 89: Wait.. I thought `isHex` already did this? Didn't we discuss this at some point?

packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
```diff
@@ -105,7 +109,7 @@ const lookupTokenIdConfigure = (
 
             let value: ContractTokenDetails[] = [];
 
-            if (!searchQuery) {
+            if (!searchQuery || !isHex(searchQuery) || !isEven(searchQuery.length)) {
```
soerenbf, line 112:
```suggestion
            if (!searchQuery || !isHex(searchQuery) || !isEven(searchQuery.length)) {
                setResult({ loading: false, value: undefined });
```

And then remove the conditional at line 179-181 (which does exactly the same, just prior to entering this function).

soerenbf (changes requested):

packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
shjortConcordium: It doesn't seem to. :shrug:

packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
shjortConcordium: :thinking: 
Why?

soerenbf: The conditional at 179-181 only runs on `!searchQuery`, but the one marked above is more lax. As such, you could end in a place where the latest `setResult` call is `{loading: true, ...}`.

Either the 2 conditionals need to be aligned, or they should be consolidated as requested in my previous comment.

packages/browser-wallet/src/popup/pages/Account/ManageTokens/TokenList.tsx
shjortConcordium: It doesn't: (The one that was referenced)
https://github.com/Concordium/concordium-node-sdk-js/blob/64baa0938fbddd813b68fdd91b012769667fe16c/packages/common/src/util.ts#L96

soerenbf (approved):

PR #201, opened 2022-11-29
Integrate WalletConnect

bisgardo:
## Purpose

Implement support for doing chain interaction via WalletConnect instead of the Browser Wallet.

## Changes

The dapp now interacts with the chain via a "wallet connection" interface that has implementations for both Browser Wallet and WalletConnect.

The interfaces and implementations are intended to be pulled out as a separate library.

The user selects the implementation to use via buttons above the "connect" button (which is displayed only once an implementation has been selected).


examples/wCCD/src/wallet/WalletConnect.ts
```diff
@@ -0,0 +1,296 @@
...
+            }
+        });
+        // this.client.on('session_update', ({ topic, params }) => {
+        //     const { namespaces } = params;
+        //     // Overwrite the `namespaces` of the existing session with the incoming one.
+        //     const updatedSession = { ...session, namespaces };
+        //     // Integrate the updated session state into your dapp state.
+        //     // eslint-disable-next-line no-console
+        //     console.debug('Wallet Connect event: session_update', { topic, updatedSession });
+        // });
```
shjortConcordium, line 285:
```suggestion
```

examples/wCCD/src/wallet/WalletConnect.ts
```diff
@@ -0,0 +1,296 @@
...
+        this.client.on('session_event', ({ topic, params: { chainId: cid, event }, id }) => {
+            // Handle session events, such as "chainChanged", "accountsChanged", etc.
+            // eslint-disable-next-line no-console
+            console.debug('Wallet Connect event: session_event', { topic, id, chainId: cid, event });
+            switch (event.name) {
+                case 'chanChanged':
+                    events.onChainChanged(event.data); // TODO implement correctly
+                    break;
+                case 'accountsChanged':
+                    events.onAccountChanged(event.data); // TODO implement correctly
```
shjortConcordium, line 271: What does these TODO imply, that this is wrong?

examples/wCCD/src/wallet/WalletConnect.ts
```diff
@@ -0,0 +1,296 @@
...
+        });
+        events.onAccountChanged(resolveAccount(session));
+
+        // Register event handlers (from official docs).
+        this.client.on('session_event', ({ topic, params: { chainId: cid, event }, id }) => {
+            // Handle session events, such as "chainChanged", "accountsChanged", etc.
+            // eslint-disable-next-line no-console
+            console.debug('Wallet Connect event: session_event', { topic, id, chainId: cid, event });
+            switch (event.name) {
+                case 'chanChanged':
```
shjortConcordium, line 267: is `chanChanged` the actual event name? :thinking:

examples/wCCD/src/constants.ts
```diff
@@ -22,3 +24,6 @@ export const VIEW_FUNCTION_RAW_SCHEMA =
...
 export const BALANCEOF_FUNCTION_RAW_SCHEMA =
     '//8CAQAAAAkAAABjaXMyX3dDQ0QBABQAAgAAAAMAAAB1cmwWAgQAAABoYXNoFQIAAAAEAAAATm9uZQIEAAAAU29tZQEBAAAAEyAAAAACAQAAAAkAAABiYWxhbmNlT2YGEAEUAAIAAAAIAAAAdG9rZW5faWQdAAcAAABhZGRyZXNzFQIAAAAHAAAAQWNjb3VudAEBAAAACwgAAABDb250cmFjdAEBAAAADBABGyUAAAAVBAAAAA4AAABJbnZhbGlkVG9rZW5JZAIRAAAASW5zdWZmaWNpZW50RnVuZHMCDAAAAFVuYXV0aG9yaXplZAIGAAAAQ3VzdG9tAQEAAAAVCQAAAAsAAABQYXJzZVBhcmFtcwIHAAAATG9nRnVsbAIMAAAATG9nTWFsZm9ybWVkAg4AAABDb250cmFjdFBhdXNlZAITAAAASW52b2tlQ29udHJhY3RFcnJvcgITAAAASW52b2tlVHJhbnNmZXJFcnJvcgIaAAAARmFpbGVkVXBncmFkZU1pc3NpbmdNb2R1bGUCHAAAAEZhaWxlZFVwZ3JhZGVNaXNzaW5nQ29udHJhY3QCJQAAAEZhaWxlZFVwZ3JhZGVVbnN1cHBvcnRlZE1vZHVsZVZlcnNpb24C';
+
+export const WALLET_CONNECT_PROJECT_ID = '76324905a70fe5c388bab46d3e0564dc';
+export const TESTNET = new Network('testnet', TESTNET_GENESIS_BLOCK_HASH, 'https://json-rpc.testnet.concordium.com');
```
DOBEN, line 29: If the user tries to connect but its browser wallet is set to mainnet, the connection button does not work. Could we add an error message, explaining to the user that it needs to switch its browser wallet to testnet to get the connection button to work.

If a user connects their wallet to testnet and then later switches to mainnet and tries to use the wCCD front-end again the following error is shown: "Site is not whitelisted". Could we add some better error message guiding the user back to what he/she has to do to be able to use the front-end again.

examples/wCCD/src/wCCD.tsx
```diff
@@ -293,31 +364,28 @@ export default function wCCD({ handleGetAccount, handleNotConnected }: Props) {
...
+                                        : unwrap(
+                                              walletConnection,
+                                              connectedAccount,
+                                              WCCD_CONTRACT_INDEX,
+                                              CONTRACT_SUB_INDEX,
+                                              amount
+                                          );
+                                    tx.then(setHash)
+                                        .catch((err) => setError((err as Error).message))
+                                        .finally(() => setWaitingForUser(false));
```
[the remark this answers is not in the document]
bisgardo: Sure why not :)

examples/wCCD/src/wCCD.tsx
```diff
@@ -101,146 +122,196 @@ async function updateWCCDBalanceAccount(account: string, setAmountAccount: (x: b
...
+                    throw new Error(`invalid connector type '${connectorType}'`);
             }
         }
-    }, [isConnected]);
+    }, [connectorType]);
 
-    const handleOnClick = useCallback(() => {
-        setWaitForUser(true);
+    const [waitingForUser, setWaitingForUser] = useState<boolean>(false);
+    const [connectedAccount, setConnectedAccount] = useState<string>();
```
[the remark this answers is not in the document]
bisgardo: This has been fixed now.

examples/wCCD/src/wCCD.tsx
```diff
@@ -101,146 +123,196 @@ async function updateWCCDBalanceAccount(account: string, setAmountAccount: (x: b
...
             <h3>Wrap and unwrap your CCDs and wCCDs on the Concordium Testnet</h3>
             <div style={blackCardStyle}>
                 <div>
-                    {!isConnected && waitForUser && (
+                    <button
+                        style={connectorTypeStyle(connectorType === 'BrowserWallet')}
+                        type="button"
+                        onClick={() => setConnectorType('BrowserWallet')}
+                    >
+                        Use Browser Wallet
```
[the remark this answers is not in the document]
bisgardo: The behavior now is that if you're on one type of connection and switch to the other, the existing connection will be terminated. I don't know if that's the best experience; maybe one should explicitly disconnect before being allowed to change connector. I would like that to be changed separately though.

It's also possible to keep connections alive even if they aren't used, allowing the user to switch back and forth (admittedly not expecting this to be a common usecase).

examples/wCCD/src/constants.ts
bisgardo: @DOBEN Are these issues introduced by this PR?

I don't have mainnet accounts in my browser wallet, but as far as I can tell there is no way to tell the network settings of the browser from the app? @shjortConcordium 

There is a handler for `chainChanged` events which I would expect to trigger when the settings change, but it doesn't seem to be the case... That handler would check the genesis hash against the expected one and raise an alert like it was before.

examples/wCCD/src/wCCD.tsx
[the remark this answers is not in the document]
bisgardo: I've changed it such that the connector button turns into "Disconnect" when there is an active account with that connector. And that you must explicitly disconnect before the other connector is enabled.

examples/wCCD/src/wallet/WalletConnect.ts
```diff
@@ -0,0 +1,296 @@
...
+        if (typeof value === 'bigint') {
+            return Number(value);
+        }
+        return value;
+    });
+}
+
+/**
+ * Encode parameters into appropriate payload field ('payload.param' for 'InitContract' and 'payload.message' for 'Update').
+ * The 'parameters' and 'schema' parameters must be not undefined for these transaction types.
```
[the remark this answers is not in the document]
shjortConcordium: :shrug: 
https://github.com/Concordium/concordium-browser-wallet/blob/22150735f9e87a1f36c37a5659e58fea660026db/examples/piggybank/src/utils.ts#L46

This is part of the sign and send transaction, for which the browser wallet does accept these as undefined.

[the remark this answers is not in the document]
bisgardo: I guess the important part is that `initContractPayload.param` is not used and don't remember why I thought the others must be set. On the other hand, didn't we agree to change it such that the app is responsible for serializing the parameters, meaning that this whole block of code goes away soon anyway? Given this, do you still think this needs to be changed now?

examples/wCCD/src/wCCD.tsx
```diff
@@ -1,13 +1,18 @@
 /* eslint-disable no-console */
```
shjortConcordium, line 1: - Why is the disconnect button darker than others, instead of the standard green?
- When I disconnect my account with the browser wallet, I get to:
Screenshot from 2022-12-13 14-41-55
I guess the connect/disconnect buttons should reset? :thinking: 

- When I disconnect with wallet-connect in the mobile wallet, nothing happens on the dApp? :thinking:
(and vice versa?)
Should they not get disconnect events? (It might be that I have an outdated wallet) :thinking:

examples/wCCD/src/wallet/WalletConnect.ts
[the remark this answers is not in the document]
shjortConcordium: No, it doesn't, it does not matter for this example.

examples/wCCD/src/constants.ts
shjortConcordium: There is an issue, where the event doesn't trigger if you don't have an account on the network that you change to, but otherwise it should trigger.

examples/wCCD/src/wCCD.tsx
```diff
@@ -101,146 +112,143 @@ async function updateWCCDBalanceAccount(account: string, setAmountAccount: (x: b
...
-                            .then((result) => {
-                                if (result === undefined || result?.value === null) {
-                                    handleNotConnected();
-                                    /* eslint-disable no-alert */
-                                    window.alert(
-                                        'Your JsonRpcClient in the Concordium browser wallet cannot connect. Check if your Concordium browser wallet is connected to testnet!'
-                                    );
-                                }
-                            })
-                    );
```
shjortConcordium, line 171: Here.

examples/wCCD/src/constants.ts
shjortConcordium: And you can at https://github.com/Concordium/concordium-browser-wallet/pull/201#discussion_r1047181296 where it previously checked for the network.

examples/wCCD/src/wCCD.tsx
bisgardo: The color was set to indicate that "Use Browser Wallet" was selected. This was before I changed it to be a disconnect button. I think there needs to be a visual indicator of which connection type is used, but am not sure if this is the best solution. Maybe if the button turned red and said "Disconnect Browser Wallet" would be better?

bisgardo: I haven't seen this error other than the mobile wallet often not picking up that the dapp disconnected. Do you get anything useful in the console?

examples/wCCD/src/wCCD.tsx
bisgardo: Yes? That check has been moved to `onChainChanged` in `WithWalletConnection`.

examples/wCCD/src/constants.ts
bisgardo: That check is still happening in `onChainChanged` in `WithWalletConnection`.

examples/wCCD/src/wCCD.tsx
DOBEN: I love the new user flow that the user has to actively disconnect before being able to switch to a new connection. Well done. I also agree that we should make the `Disconnect` button red and call it `Disconnect Browser Wallet/Disconnect Wallet Connect`.

When disconnecting in the browser wallet then this is something the front-end cannot pick up it seems. If we make the `wrap/unwrap` button grey in that case, the user would know that it needs to `disconnect` (only button left to click on dApp front-end) the connection on the dApp as well.
Screenshot from 2022-12-14 11-45-38

examples/wCCD/src/constants.ts
DOBEN: I checked again and as @shjortConcordium described the edge case happens if the user is on mainnet (with no mainnet account) then the `Connect Browser Wallet` button does not work (and we have no error message that would hint to the user to switch to testnet). A similar edge case is when the user switches to mainnet  (with no mainnet account) while already connected to the side. 

Feel free to resolve this thread if you think that we would first have to add some events/getter functions in the wallet that would support this feature maybe.

DOBEN (approved):
Well done.

examples/wCCD/src/wCCD.tsx
bisgardo: Updated styling of both disconnect and styling buttons as you suggested.

examples/wCCD/src/constants.ts
bisgardo: I filed https://concordium.atlassian.net/browse/CBW-633 with suggestions to improve the BW to improve this situation. I think the remaining issues are edge case annoyances that we can live with for a short while, so I'll merge this now and look into improving the error message in a followup. This allows us to get this feature out and make managers happy again :)

I'll also address a minor issue that the BW doesn't really support disconnecting, so when we "do" that the event handlers hang on and get reregistered on a "new" connection. This causes them to fire as many times as the connection has been reset.

@shjortConcordium If you continue to see connection issues, please report it as an issue or comment it on the next PR.

PR #203, opened 2022-12-04
Handle account limit when picking identity to create a new account with

shjortConcordium:
## Purpose

Block user from trying (and failing) to create accounts on identities which has already reached the limit.

## Changes

- Disable identity cards in list when creating new account, if limit has been reached. (and display a message)
- IdCard now have pointer cursor when clickable

## Checklist

      hard-to-understand areas.


soerenbf (approved):
Looks good

PR #207, opened 2022-12-06
Re-arrange NFT details

soerenbf:
## Purpose

We want the NFT image to be displayed at the top of the NFT details page.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/shared/TokenDetails/TokenDetails.tsx
+++ b/packages/browser-wallet/src/popup/shared/TokenDetails/TokenDetails.tsx
@@ -86,6 +86,11 @@ function Nft({ token, balance, contractIndex, subIndex }: TokenProps) {
                 {thumbnail && <Img src={thumbnail.url} alt={name} withDefaults />}
                 {name}
             </h3>
+            {display?.url && (
+                <div className="token-details__image">
+                    <img src={display?.url} alt={name} />
+                </div>
+            )}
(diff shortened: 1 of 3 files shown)


shjortConcordium (commented):
.

packages/browser-wallet/src/popup/shared/TokenDetails/TokenDetails.tsx
```diff
@@ -86,6 +86,11 @@ function Nft({ token, balance, contractIndex, subIndex }: TokenProps) {
                 {thumbnail && <Img src={thumbnail.url} alt={name} withDefaults />}
                 {name}
             </h3>
+            {display?.url && (
```
shjortConcordium, line 89: There is no limit on the height of the image? Isn't that an issue?

soerenbf: I don't think so.. What's worse, to scroll a lot to see the full image, or to not be able to see it due to it being slimmed down to nothing due to a height limit? I prefer the scrolling option, I think? :shrug:

shjortConcordium: I was more thinking about the other data. I guess I just disagree with it being on top :thinking:

shjortConcordium (approved):

PR #208, opened 2022-12-06
Change transaction status instead of removing them

soerenbf:
## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/store/transactions.ts
+++ b/packages/browser-wallet/src/popup/store/transactions.ts
@@ -19,27 +19,39 @@ const monitorTransactionStatus = (genesisHash: string) => {
     monitoredMap[genesisHash] = monitoredMap[genesisHash] ?? [];
 
     /**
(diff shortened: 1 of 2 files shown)


shjortConcordium (approved):

packages/browser-wallet/src/popup/store/transactions.ts
```diff
@@ -19,27 +19,39 @@ const monitorTransactionStatus = (genesisHash: string) => {
     monitoredMap[genesisHash] = monitoredMap[genesisHash] ?? [];
 
     /**
-     * Resolves with true if monitoring has started with this function invocation and false if transaction is already being monitored.
+     * Resolves with {status: TransactionCost; cost: bigint} if monitoring has started with this function invocation and false if transaction is already being monitored.
```
soerenbf, line 22:
```suggestion
     * Resolves with {status: TransactionStatus; cost: bigint} if monitoring has started with this function invocation and false if transaction is already being monitored.
```

PR #211, opened 2022-12-07
Update account info on address subscribe invocation

soerenbf:
## Purpose

Immediately get new account info when opening a view showing account info

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/shared/account-info-listener.ts
+++ b/packages/browser-wallet/src/popup/shared/account-info-listener.ts
@@ -23,7 +23,30 @@ export class AccountInfoListener extends EventEmitter {
         this.genesisHash = network.genesisHash;
     }
 
+    private async update(accountAddress: string, blockHash?: string) {
+        try {
+            if (this.accountsMap.size > 0) {
+                const bh = blockHash ?? (await this.client.getConsensusStatus()).lastFinalizedBlock;
+                const address = new AccountAddress(accountAddress);
+                const accountInfo = await this.client.getAccountInfo(address, bh);
+
+                if (accountInfo) {
+                    updateRecord(
+                        accountInfoCacheLock,
+                        useIndexedStorage(sessionAccountInfoCache, async () => this.genesisHash),
+                        accountInfo.accountAddress,
+                        stringify(accountInfo)
+                    );
+                }
+            }
+        } catch (e) {
+            this.emit('error', e);
+        }
+    }
+
     subscribe(accountAddress: string) {
+        this.update(accountAddress);
+
         const existingSubscriptions = this.accountsMap.get(accountAddress);
         if (existingSubscriptions !== undefined) {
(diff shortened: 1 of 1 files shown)


packages/browser-wallet/src/popup/shared/account-info-listener.ts
```diff
@@ -23,7 +23,30 @@ export class AccountInfoListener extends EventEmitter {
         this.genesisHash = network.genesisHash;
     }
 
+    private async update(accountAddress: string, blockHash?: string) {
+        try {
+            if (this.accountsMap.size > 0) {
```
shjortConcordium, line 28: I do not think this check should be in here. (Why did you move it inside the for loop?) :thinking:

shjortConcordium:
I'm not sure this is a good solution. I don't think it should reload the every account info every time the list is opened. (and on every interval while it is open)

soerenbf:
> I'm not sure this is a good solution. I don't think it should reload every account info every time the list is opened. (and on every interval while it is open) What I expected for this was to trigger an update, if you resolved a transfer sent to another account in the wallet. thinking

The current solution on main subscribes all accounts in the account list to be updated with the already ongoing interval. This way, it will seem completely random if accounts in the account list, and even the account balance view, or anywhere else we display account information, are updated or not. I think the more consistent approach is to update them all immediately + add the to the interval, which is what I attempted to do here.

Maybe the proposed solution is too aggressive, but I think the current solution is bad user experience.

Why don't you want the account list to be updated every time you open an account? Isn't it a fair assumption that users would like to see an updated account balance immediately instead of within 0-15 seconds on random?

orhoj:
@soerenbf and @shjortConcordium 

It seems like this change has stranded. Is it still relevant and do we need to discuss whether it is a relevant change or not?

soerenbf:
The underlying issue has been fixed elsewhere, but what this implements is updating accounts with the current balance on chain immediately instead of at a random interval. The discussion in the comments pertain to whether updating all accounts (i.e. also when opening the account list) is too aggressive or not.

As I remember, this is an issue both on the account list, but also on the account page. At least opening the account page for an account should trigger an update immediately for that account.

orhoj:
> The underlying issue has been fixed elsewhere, but what this implements on top, is updating accounts with the current balance on chain immediately instead of at a random interval. The discussion in the comments pertain to whether updating all accounts (i.e. also when opening the account list) is too aggressive or not.
> 
> As I remember, this is an issue both on the account list, but also on the account page. At least opening the account page for an account should trigger an update immediately for that account.

It is too aggressive to update everything in the account list when opening it. This is an issue in the desktop wallet, where we have seen users with many accounts hitting the rate limit set by VirtualHive.

If you think it would still be nice to have the account information updated when selecting a particular account, then please make a ticket for that so we can handle that separately from this pull request.

PR #214, opened 2022-12-08
Id2 view alignments

soerenbf:
## Purpose

Various UI alignments for ID proof view.

## Changes

- Aligned external request layout header
- Show account/identity included in message payload instead of selected in external request layout
- Improve loading view when generating proof
- Removed statement line separator

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/store/identity.ts
+++ b/packages/browser-wallet/src/popup/store/identity.ts
@@ -65,26 +65,28 @@ export const setRecoveryPayloadAtom = atom<null, RecoveryPayload, Promise<void>>
     set(recoveryStatusAtom, { payload })
 );
 
-export const identityByAddressAtomFamily = atomFamily<string, Atom<AsyncWrapper<ConfirmedIdentity | undefined>>>(
-    (address) =>
-        atom((get) => {
-            const creds = get(credentialsAtomWithLoading);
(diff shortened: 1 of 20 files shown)


packages/browser-wallet/src/popup/store/identity.ts
```diff
@@ -65,26 +65,28 @@ export const setRecoveryPayloadAtom = atom<null, RecoveryPayload, Promise<void>>
...
+    atom((get) => {
+        const creds = get(credentialsAtomWithLoading);
+        const ids = get(identitiesAtomWithLoading);
 
-            if (creds.loading || ids.loading) {
-                return { loading: true, value: undefined };
-            }
+        if (creds.loading || ids.loading) {
+            return { loading: true, value: undefined };
+        }
```
shjortConcordium, line 78: Is it just formatting of the first line that caused the change in indent? (Can't find another reason)

shjortConcordium (approved):

packages/browser-wallet/src/popup/store/identity.ts
soerenbf: it's the change in the type from `string` to `string | undefined` in the first type param of `atomFamily`

PR #215, opened 2022-12-09
Id card alignments

soerenbf:
## Purpose

Align ID card + attributes views with newest iteration from mobile wallet design.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
```diff
@@ -100,12 +121,21 @@ export default function IdCard({ name, provider, status, onNameChange, className
             }}
         >
             <header className="id-card__header">
```
shjortConcordium, line 123: Screenshot from 2022-12-09 13-15-36
Sometimes there is space between the header and the border (I think I've also seen it with the old version, though)

packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
```diff
@@ -100,12 +121,21 @@ export default function IdCard({ name, provider, status, onNameChange, className
...
+                        onClick={() => setIsEditing(true)}
+                    >
+                        <EditIcon />
+                    </Button>
+                )}
             </header>
             <div className="id-card__name">
-                {!onNameChange && <div className="id-card__name-form">{name}</div>}
-                {onNameChange && <EditableName name={name} onChange={onNameChange} />}
+                <EditableName name={name} onChange={handleNameChange} isEditing={isEditing} />
```
shjortConcordium, line 138: The checkmark feels a bit weird after the text instead of replacing the edit button :shrug:

shjortConcordium (approved):
LGTM, have some minor thoughts though:

packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
soerenbf: Yeah I agere.. I'll go over it with @jens-concordium

packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
soerenbf: You're right, fixed this. It was due to a border-width being set with rem, which doesn't seem to be a good idea in cases like this.

packages/browser-wallet/src/popup/shared/IdCard/IdCard.tsx
soerenbf: Moved the checkmark to the location of the edit button.

PR #224, opened 2022-12-14
Always use mainnet proxy for terms and conditions

shjortConcordium:
## Purpose

To remove the burden of keeping the wallet-proxy versions in sync (or risk extra prompts), we will only use the mainnet one.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/TermsAndConditions/AcceptTerms.tsx
+++ b/packages/browser-wallet/src/popup/pages/TermsAndConditions/AcceptTerms.tsx
@@ -8,8 +8,8 @@ import FormCheckbox from '@popup/shared/Form/Checkbox';
 import ExternalLink from '@popup/shared/ExternalLink';
 import urls from '@shared/constants/url';
 import Submit from '@popup/shared/Form/Submit';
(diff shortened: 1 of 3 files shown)


packages/browser-wallet/src/popup/pages/TermsAndConditions/AcceptTerms.tsx
```diff
@@ -25,8 +25,7 @@ type Props = {
 export default function AcceptTerms({ children, onSubmit }: Props) {
     const { t } = useTranslation('termsAndConditions');
     const [{ loading, value: acceptedTerms }, setAcceptedTerms] = useAtom(acceptedTermsAtom);
-    const network = useAtomValue(networkConfigurationAtom);
-    const config = useAsyncMemo(() => getTermsAndConditionsConfig(network.explorerUrl), undefined, []);
+    const config = useAsyncMemo(() => getTermsAndConditionsConfig(), undefined, []);
```
soerenbf, line 28: could just do:

```suggestion
    const config = useAsyncMemo(getTermsAndConditionsConfig, undefined, []);
```

soerenbf (approved):

PR #232, opened 2022-12-16
wCCD: Improve error handling

bisgardo:
Display error messages in the application instead of logging them to console or showing alert popups. This adds a bunch of error fields to the various components' state.

Error messages got a pass and the connection logic was migrated from the app component into the `WithWalletConnection` library component.

This improves the UX and solves the following issues:

- Browser Wallet:

  - Clean up event handlers when "disconnecting" (i.e. discarding) the connector. This avoids interference between reconnects.
  - Show error message when browser wallet isn't available.

- WalletConnect:

  - Disconnect is now detected properly.
  - Ensure that we no longer get stuck in "waiting for user" state when the modal is closed without connecting.

- RPC errors are improved and displayed in the UI. This makes it as clear as it can be if the connected account belongs to the wrong chain.

The UI now only shows the "transaction status" and "admin instance" fields when the is a connection as there otherwise isn't anything useful to show in them. As part of this the "lean how to make a wrapper like this" link was moved down next to the version.

### DIFF
--- a/examples/wCCD/src/wallet/WithWalletConnection.tsx
+++ b/examples/wCCD/src/wallet/WithWalletConnection.tsx
@@ -17,10 +17,11 @@ import { WalletConnectConnector } from './WalletConnect';
 interface State {
     activeConnectorType: string | undefined;
     activeConnector: WalletConnector | undefined;
+    activeConnectorError: string;
     activeConnection: WalletConnection | undefined;
+    // activeConnectionError: string | undefined;
     activeConnectionGenesisHash: string | undefined;
     activeConnectedAccount: string | undefined;
-    connectorError: string;
 }
 
 // TODO React appropriately if 'network' changes.
@@ -35,7 +36,6 @@ export interface WalletConnectionProps extends State {
     activeConnectionGenesisHash: string | undefined;
     setActiveConnectorType: (t: string | undefined) => void;
     setActiveConnection: (c: WalletConnection | undefined) => void;
-    connectorError: string;
 }
 
 // eslint-disable-next-line react/prefer-stateless-function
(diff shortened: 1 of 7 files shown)


soerenbf (approved):
Looks pretty good to me, and I like the changes made to the UI!

One thing I would add (but might be out of scope), is that when wrapping CCD, it would be cool if there was a popup that also asked if the user wants to add the wCCD token to their wallet, leveraging the wallet API to do so.

This is common practice on f.x. Uniswap whenever you swap tokens from one to another.

examples/wCCD/src/wallet/WithWalletConnection.tsx
```diff
@@ -112,20 +131,15 @@ export class WithWalletConnection extends React.Component<Props, State> implemen
...
+        console.debug("WithWalletConnection: calling 'onChainChanged'", { connection, genesisHash, state: this.state });
+        const { activeConnection } = this.state;
+        if (connection === activeConnection) {
+            this.setState({ activeConnectionGenesisHash: genesisHash });
         }
     };
 
     onDisconnect = (connection: WalletConnection) => {
-        console.debug('WithWalletConnection: onDisconnect called', { connection, state: this.state });
+        console.debug("WithWalletConnection: calling 'onDisconnect'", { connection, state: this.state });
```
soerenbf, line 142: Can we remove the logs before this is deployed to the production testnet site? If this is intended to be published when merged, I think we should remove them now.

soerenbf (changes requested):

examples/wCCD/src/wallet/WithWalletConnection.tsx
bisgardo: I was intending to leave them in to be able to debug issues that inevitably happen. Why would you like to remove them? It's low volume logging, mostly on the debug level.

I'll merge it like this but not deploy for now.

bisgardo:
> One thing I would add (but might be out of scope), is that when wrapping CCD, it would be cool if there was a popup that also asked if the user wants to add the wCCD token to their wallet, leveraging the wallet API to do so.

Nice idea, but yeah not in scope of this PR.

> Haven't tested the wallet connect integration, as I don't have an android phone, so if you want feedback on that, I would wait for a review from someone who can test this also.

I think I've tested it enough myself to be confident that it works :)

PR #243, opened 2023-01-10
Add eSealing front-end example

DOBEN:
## Purpose

Add an eSealing front-end example

Corresponding smart contracts:
https://github.com/Concordium/concordium-rust-smart-contracts/issues/197
https://github.com/Concordium/concordium-rust-smart-contracts/pull/207

The front-end should support the following two flows with the browser wallet (or wallet connect):

- Upload a file from the computer => register its file hash in the smart contract
Similar to:
http://filesealing.newtrust.co.s3-website-us-east-1.amazonaws.com/

- Upload a file from the computer => retrieve the time-stamp and witness (sender_account) that registered the file hash
Similar to:
https://timestamp.northstake.dk/

## Checklist

      hard-to-understand areas.


examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+    useEffect(() => {
+        // View file record.
+        if (activeConnection && activeConnectedAccount) {
+            withJsonRpcClient(activeConnection, (rpcClient) => viewFile(rpcClient, fileHash))
+                .then((record) => {
+                    setGetFileError('');
+                    setTimestamp(record.timestamp);
+                    setWitness(record.witness);
+                })
+                .catch((e) => setGetFileError((e as Error).message));
```
bisgardo, line 150: Shouldn't timestamp and witness be reset here?

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+                            }}
+                        >
+                            {activeConnectedAccount}
+                        </button>
+                        <br />
+                        <br />
+                        <div className="containerSpaceBetween">
+                            <button
+                                style={
+                                    activeConnectedAccount === undefined || isRegisterFilePage === false
```
bisgardo, line 222:
```suggestion
                                    !isRegisterFilePage
```

`activeConnectedAccount` is never undefined in this block and `isRegisterFilePage` is always a bool.

This also applies below (also the next button).

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+                            className="input"
+                            style={InputFieldStyle}
+                            type="file"
+                            onChange={changeHandler}
+                            ref={file}
+                        />
+                        <button
+                            style={ButtonStyle}
+                            type="button"
+                            onClick={async () => {
```
bisgardo, line 300: I think you should use an explicit promise here as otherwise any exception will not get caught. Maybe surrounding with try/catch is sufficient.

bisgardo (commented):
FYI The wallet connection stuff is being put in a library at https://github.com/Concordium/concordium-dapp-libraries/tree/main/packages/wallet-connectors and is almost ready for release.

It's a drop-in replacement for the classes you're using here or at least should be close to it. I'll make sure that the dapps are migrated once the library is released.

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+        activeConnector,
+        isActiveConnectorWaitingForUser,
+        activeConnection,
+        activeConnectionGenesisHash,
+        activeConnectedAccount,
+        activeConnectorError,
+        connect,
+    } = props;
+
+    const [loading, setLoading] = useState('');
```
bisgardo, line 122: I would make this a boolean `isLoading` rather than a string.

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+                                    const newFileHashHex = sha256(byteArray.toString());
+                                    setFileHash(newFileHash.toString());
+                                    setFileHashHex(newFileHashHex);
+                                    setLoading('');
+                                } else {
+                                    alert('Choose a file before uploading');
+                                }
+                            }}
+                        >
+                            Upload File
```
bisgardo, line 316:
```suggestion
                            Compute File Hash
```

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+                                onClick={() => {
+                                    setIsRegisterFilePage(true);
+                                    setFileHash('');
+                                    setFileHashHex('');
+                                    setWitness('');
+                                    setTimestamp('');
+                                    setHash('');
+                                }}
+                            >
+                                Register File Hash
```
bisgardo, line 237: It took me a moment to understand that this button doesn't actually register the hash, only selects that function. And it's a little confusing that there is another button with the same text.

Maybe make it a `<label>` instead?

bisgardo (commented):
Looks good!

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,413 @@
...
+        toBuffer(E_SEALING_RAW_SCHEMA, 'base64'),
+        E_SEALING_CONTRACT_NAME,
+        'getFile',
+        2
+    );
+
+    if (returnValues.Some !== undefined) {
+        return returnValues.Some[0];
+    }
+    return { timestamp: 'Not registered', witness: 'Not registered' };
```
abizjak, line 106: Is it intended to have the string `Not registered` as a value of a field called `timestamp`? Instead of using `null` or some such?

abizjak (approved):

examples/eSealing/README.md
```diff
@@ -0,0 +1,55 @@
...
+To build the docker image run the following:
+
+```
+docker build -t eSealing_front_end:$PROJECT_VERSION .
+```
+
+e.g.
+
+```
+docker build -t eSealing_front_end:3.0.0 .
```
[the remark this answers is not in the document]
bisgardo: You can also use
```
npm pkg get version
```
which will surely work on the dev's machine. Unfortunately, it only supports outputting in JSON, but it's still easier to read than package.json.

I would personally prefer the `jq` command with a single line of explanation and a reference to its website. But you can also just give a generic tag.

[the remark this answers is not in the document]
DOBEN: I have kept the examples with fixed versions but added a section explaining where to get the version number from (with several example commands).

examples/eSealing/src/eSealing.tsx
DOBEN: Can you elaborate on how I should use the `<label>` tag here?

I changed the description of the buttons to `Registration Tab` and `Display Tab` and also changed the colors to improve the understanding of this switch.

examples/eSealing/README.md
```diff
@@ -0,0 +1,55 @@
+# eSealing web application
+
+The example project included in this repository serves as a working example of how to integrate with smart contracts on the Concordium blockchain. This web app supports the following two flows with the browser wallet (or wallet connect):
+
+-   Upload a file from the computer => register its file hash in the smart contract
+-   Upload a file from the computer => retrieve the time-stamp and witness (sender_account) that registered the file hash
```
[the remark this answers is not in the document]
DOBEN: I changed it to:

`Compute the file hash of a selected file`

examples/eSealing/src/eSealing.tsx
DOBEN: I changed it to using `null` here.

examples/eSealing/src/utils.ts
```diff
@@ -0,0 +1,44 @@
+/* @typescript-eslint/ban-ts-comment */
+
```
bisgardo, line 2:
```suggestion
```

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,435 @@
...
+        activeConnector,
+        isActiveConnectorWaitingForUser,
+        activeConnection,
+        activeConnectionGenesisHash,
+        activeConnectedAccount,
+        activeConnectorError,
+        connect,
+    } = props;
+
+    const [isLoading, setLoading] = useState<boolean>(false);
```
bisgardo, line 146:
```suggestion
    const [isLoading, setLoading] = useState(false);
```

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,435 @@
...
+                                        const arrayBuffer = await selectedFile.arrayBuffer();
+                                        const byteArray = new Uint8Array(arrayBuffer as ArrayBuffer);
+                                        const newFileHashHex = sha256(byteArray.toString());
+                                        setFileHashHex(newFileHashHex);
+                                        setLoading(false);
+                                    } else {
+                                        alert('Choose a file to compute the file hash');
+                                    }
+                                } catch (err) {
+                                    console.error(setLoadingError((err as Error).message));
```
bisgardo, line 330:
```suggestion
                                    setLoadingError((err as Error).message);
```

Setter returns void.

bisgardo (approved):

examples/eSealing/src/eSealing.tsx
bisgardo: I was just thinking of rendering it as text instead of a button along the lines of how `<label>` is used for form fields. I don't know whether `<label>` can switch the toggle natively the same way here, but it might at least give you the "correct" styling for free. You can always use the same click handler like you do now, regardless of what HTML element you use.

But I don't think you need to change it, the current solution is fine with me.

examples/eSealing/src/eSealing.tsx
```diff
@@ -0,0 +1,435 @@
...
+            <div>
+                <br />
+                Version: {version} |{' '}
+                <a
+                    style={{ color: 'white' }}
+                    href="https://developer.concordium.software/en/mainnet/smart-contracts/tutorials/wCCD/index.html"
+                    target="_blank"
+                    rel="noreferrer"
+                >
+                    Learn how to make a wrapper like this
```
bisgardo, line 429: Is this a wrapper?

DOBEN: I will update this text and this link to the written eSealing documentation now:

https://github.com/Concordium/concordium.github.io/pull/727

PR #247, opened 2023-01-16
CI: Include hash of package.json in cache key

bisgardo:
## Purpose

Ensure that changes in library functions that didn't involve dependency changes invalidate the CI's dependency cache.

## Changes

Include hash of `package.json` in the cache key.

Also included a dummy value for force invalidating the cache and parameterize node version.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -10,6 +10,10 @@ on:
     # Allows us to run the workflow manually from the Actions tab
     workflow_dispatch:
 
+env:
+    dummy: 1 # change to force cache invalidation
+    node_version: 14.16.0
+
 jobs:
     dependencies:
         runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


bisgardo:
@DOBEN the problem is that the wallet API changed in https://github.com/Concordium/concordium-browser-wallet/commit/f3a5e871b53725c46fe4112b91f59e2786e0feb8 without a version bump. So because your project references the same version of the API as the one specified in the repo, it will use that instead of downloading the published package.

The easiest solution is to migrate to the new type.

I don't know how to avoid this in the future. @shjortConcordium any thoughts? I guess it comes down to what "version" means in between releases.

DOBEN (approved):

shjortConcordium:
> I don't know how to avoid this in the future. @shjortConcordium any thoughts? I guess it comes down to what "version" means in between releases.

One thing would be to bump the api version when doing changes, but I think the correct thing would be to move all these examples into a separate repository. (Which also makes sense now that they no longer only used with the browser wallet)

abizjak:
> > I don't know how to avoid this in the future. @shjortConcordium any thoughts? I guess it comes down to what "version" means in between releases.
> 
> One thing would be to bump the api version when doing changes, but I think the correct thing would be to move all these examples into a separate repository. (Which also makes sense now that they no longer only used with the browser wallet)

There are external links to the examples already, from developer documentation, and also we have given them out to various people. So please don't move stuff without addressing this issue.

shjortConcordium:
> > > I don't know how to avoid this in the future. @shjortConcordium any thoughts? I guess it comes down to what "version" means in between releases.
> > 
> > 
> > One thing would be to bump the api version when doing changes, but I think the correct thing would be to move all these examples into a separate repository. (Which also makes sense now that they no longer only used with the browser wallet)
> 
> There are external links to the examples already, from developer documentation, and also we have given them out to various people. So please don't move stuff without addressing this issue.

I'm not saying to rush moving them, but I do think they should be moved. But when we would do it, I would update the documentation and leave the examples here as readme's with links to the new location, to avoid the broken links.
It might not be relevant here though, because it's just that the api version should be bumped/the examples should be updated to work with the changes.

PR #249, opened 2023-01-20
eSealing: Fix label and error resetting

bisgardo:
The label around the file input covered a lot of elements below the component, such that e.g. clicking on the computed file hash opened the file dialog.

Also, once a file hash computation returned an error, this error would never be reset even if it was successful afterwards (can trigger error on Ubuntu by selecting a text file, editing the file, then computing the hash). Fixed by resetting the error on success.

The change is nowhere near as complex as GitHub makes it look. A proper viewer will reveal that most of it is reindentation.

### DIFF
--- a/examples/eSealing/package.json
+++ b/examples/eSealing/package.json
@@ -1,7 +1,7 @@
 {
     "name": "e_sealing",
     "packageManager": "yarn@3.2.0",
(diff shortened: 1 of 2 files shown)


DOBEN (approved):
You could increase the minor version in the eSealing front-end example.

PR #250, opened 2023-02-09
Migrate example dApps to the newest version of dapp-libraries

bisgardo:
## Purpose

Ensure that the example dApps use the newest version of dapp-libraries.

## Changes

Bump the dependencies to dapp-libraries to v0.2.0 for wCCD and e-Sealing.


DOBEN (approved):
Looks good. I also tried the wCCD and eSealing front-end with these wallet-connect changes and works as expected.

Issue #252, opened 2023-02-15
Token Metadata Deserialization Error


parv0888:
**Bug Description**
Token Metadata De-serialization error in the method
https://github.com/Concordium/concordium-browser-wallet/blob/main/packages/browser-wallet/src/shared/utils/token-helpers.ts#L61

The Method linked above should also take into account the optional Hash value which can be present in the [Token Metadata](https://proposals.concordium.software/CIS/cis-2.html#metadataurl)

**Steps to Reproduce**
Using the URL
https://chainorders.github.io/concordium-nft-tutorials/index.html#/mint-multi-batch
1. Click "Deploy New" (This will initialise a new CIS2-Multi contract).
2. Enter Pinata JWT and Click "Connect"
3. Click Upload and Select at-least 3 Image Files
4. Click "Set Token Id" in all the 3 boxes
5. Click "Upload Image" in all the 3 boxes
6. Click "Create" On all the 3 boxes
7. Set Any Quantity in all the 3 boxes
8. Click "Done" at the bottom
9. Click "Mint"
10. Upon See the Alert Message "All Minted"
11. Open the Wallet and try to "Look for Tokens" in the recently initialised contract. 
You should see the error in the wallet console. 

Alternatively You can use the contract [present here](https://github.com/chainorders/concordium-nft-tutorials/blob/main/cis2-multi/src/lib.rs). Initialize it and mint atleast 3 Tokens with Metadata `Hash`

**Expected Result**
Tokens should be added to the wallet

**Actual Result**
Error in console and wallet doesnt appear to be doing anything
```
Uncaught (in promise) RangeError: Trying to access beyond buffer length
    at _B (index.js:1127:36)
    at WA.readUint16LE.WA.readUInt16LE (index.js:1173:18)
    at QYA (token-helpers.ts:69:28)
    at token-helpers.ts:125:29
```

**Versions**
- Software Version : 0.9.6
- OS : Ubuntu
- Browser : Chrome
- Mobile device, if applicable

PR #254, opened 2023-02-16
Fix token metadata url parsing with checksum

shjortConcordium:
## Purpose

Fixes https://github.com/Concordium/concordium-browser-wallet/issues/252

## Changes

- Fix deserialization.
- validate content against checksum, if present.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/shared/utils/token-helpers.ts
+++ b/packages/browser-wallet/src/shared/utils/token-helpers.ts
@@ -3,11 +3,11 @@ import uleb128 from 'leb128/unsigned';
 import {
     AccountAddress,
     CcdAmount,
-    HexString,
     InstanceInfo,
     JsonRpcClient,
     serializeUpdateContractParameters,
     UpdateContractPayload,
+    sha256,
 } from '@concordium/web-sdk';
 import { MetadataUrl, NetworkConfiguration, TokenMetadata, TokenIdAndMetadata } from '@shared/storage/types';
 import { CIS2_SCHEMA_CONTRACT_NAME, CIS2_SCHEMA } from '@popup/constants/schema';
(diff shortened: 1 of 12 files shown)


packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -147,12 +164,23 @@ function confirmString(field?: any) {
...
  * Fetches token metadata from the given url
  */
-export async function getTokenMetadata(tokenUrl: string, network: NetworkConfiguration): Promise<TokenMetadata> {
-    if (!isMainnet(network) && tokenUrl.includes('https://some.example/token/')) {
-        const id = tokenUrl.split('https://some.example/token/')[1]?.toLowerCase() ?? 'fallback';
+export async function getTokenMetadata(
+    { url, hash: checksumHash }: MetadataUrl,
+    network: NetworkConfiguration
+): Promise<TokenMetadata> {
+    if (!isMainnet(network) && url.includes('https://some.example/token/')) {
```
abizjak, line 182: I see this was already there before, but can you explain what the purpose of this function is? It seems awfully ad-hoc behaviour.

abizjak (approved):

packages/browser-wallet/src/shared/utils/token-helpers.ts
shjortConcordium: Yes, this is very ad-hoc, I think it was mostly for development, so that we could add example tokens without actually hosting the metadata. (And so the old wCCD actually worked, it had `https://some.example/token/wccd` as metadata url)
But maybe now is the time to remove it, as it could confuse/mislead "developers" testing their tokens on testnet (if they _are_ broken)

abizjak: Yeah. It is always better to fail I think then trying to make it work in an ad-hoc manner that works some of the time. So will you remove this?


I realize that's not the Javascript way, but it should at least be our way :)

PR #260, opened 2023-03-08
WIP: Add sponsored tx example

DOBEN:
## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

_Describe the changes that were needed.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


examples/sponsoredTXs/src/constants.ts
```diff
@@ -0,0 +1,45 @@
...
+export const CONTRACT_SUB_INDEX = 0n;
+
+const WALLET_CONNECT_PROJECT_ID = '76324905a70fe5c388bab46d3e0564dc';
```
abizjak, line 25: Where does this come from?

examples/sponsoredTXs/src/sponsoredTx.tsx
```diff
@@ -0,0 +1,949 @@
...
+                                </>
+                            )}
+                            <label>
+                                <p style={{ marginBottom: 0 }}>Signer:</p>
+                                <input
+                                    className="input"
+                                    style={InputFieldStyle}
+                                    id="signer"
+                                    type="text"
+                                    placeholder="4HoVMVsj6TwJr6B5krP5fW9qM4pbo6crVyrr7N95t2UQDrv1fq"
```
abizjak, line 813: What is this account?

examples/sponsoredTXs/src/utils.ts
```diff
@@ -0,0 +1,346 @@
...
+            index: Number(SPONSORED_TX_CONTRACT_INDEX),
+            subindex: 0,
+        },
+        entry_point: 'contract_transfer',
+        nonce: Number(nonce),
+        payload: {
+            Transfer: [
+                [
+                    {
+                        amount: '1',
```
abizjak, line 136: What is the meaning of this hardcoded amount?

abizjak (commented):
First, good job reverse engineering the wallet implementation, which ideally should not be necessary.

I did not look into all the details, but regarding this example, I am not sure I understand how it is meant to work/how real it is.

As I see it here both the signing the message and signing the transaction is done via the browser wallet, right?

The use-case of sponsored transactions, as I understand it, is that the user does the inner signing, and then sends the transaction to some other service (i.e., backend) which then signs the transaction and sends it to the chain. 

I am not sure we can live with everything being done here in the browser, since it is not a representative example.

I also think this example, since it will likely include a backend, should live in dapp-examples repository, not here (like the gallery we have there).

examples/sponsoredTXs/src/constants.ts
DOBEN: https://concordium.slack.com/archives/C037WKK99NE/p1679042051884009?thread_ts=1679041309.052459&cid=C037WKK99NE

The  WALLET_CONNECT_PROJECT_ID, is the ID allocated to the account that Concordium has with walletconnect. Other dApp developers should use the same number. It is probably just for grouping traffic or something. It is the ID for our wallets, so WC knows which app to open a connection with.

examples/sponsoredTXs/src/sponsoredTx.tsx
DOBEN: This is a random account from me, so people see in which format they should provide the account.

examples/sponsoredTXs/src/utils.ts
DOBEN: The smart contract is an NFT smart contract. So very every `token_ID` there exists exactly one token. Transfer this token means that the amount is always 1.

examples/sponsoredTXs/src/constants.ts
abizjak: Should this not be a constant defined in our libraries then?

Can you make an issue to have this added?

DOBEN: https://github.com/Concordium/concordium-dapp-libraries/issues/30

DOBEN:
Moving this MR to:

https://github.com/Concordium/concordium-dapp-examples/pull/6

PR #261, opened 2023-03-13
Fix token decimals validation and crash when token integer value is above 64 bits

shjortConcordium:
## Purpose

Fix bugs related to tokens, that could hinder the usage of bridge token's.

## Changes

- Fix validation for tokens being hardcoded to assume 6 decimals.
- Fix transferring tokens with integer value above 64 bits causing a crash.
- Added tests for the issues (and test setup)

## Checklist

      hard-to-understand areas.


soerenbf (approved):
The test file should probably be called `create-transfer.test.ts` to align with other test files though.

Nice with unit tests 👍

PR #264, opened 2023-03-24
Dapp chain awareness

soerenbf:
## Purpose

Solves the issue of dApps in some cases not getting notified when the chain selected in the wallet changes.

## Changes

- Propagate `chainChanged` event to all listeners
- Emit `accountDisconnected` event if selected account on newly selected chain is not whitelisted
- Expose entrypoint in wallet API for getting the genesis hash of the chain selected in the wallet. 

## Checklist

      hard-to-understand areas.


packages/browser-wallet-api-helpers/README.md
```diff
@@ -74,6 +74,15 @@ if (accountAddress) {
 }
 ```
 
+### getSelectedChain
```
abizjak, line 77: can we make a task to make use of this API in the wCCD and other examples. Right now there is a hacky solution there.

packages/browser-wallet/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 -   Use gRPC-web instead of json-RPC.
 
+### Fixed
+
+-   `chainChanged` event is correctly propagated to all (not just whitelisted) dapps listening for events through the wallet API.
```
abizjak, line 11: Is this something that we actually want? 

The way I understand this change is that if I just have a site opened without ever connecting to the wallet I would be able to listen to these events. Is that correct?

What information do I get from this event? Nothing more than the chain genesis has changed?

abizjak (commented):
I just have some questions.

packages/browser-wallet/CHANGELOG.md
soerenbf: Yes, I think this should be up for discussion as well. The only information emitted with the event is the genesis hash of the chain the wallet has been set to use.

I couldn't come up with any scenarios where this could be used for anything malicious, but might be some sort of "invasion of privacy" anyway?

packages/browser-wallet-api-helpers/README.md
soerenbf: Sure, I'll put an issue in the backlog.

packages/browser-wallet/src/popup/store/account.ts
```diff
@@ -24,7 +24,13 @@ export const selectedAccountAtom = atom<string | undefined, string | undefined>(
     (get) => get(storedAccountAtom),
     (_, set, address) => {
         set(storedAccountAtom, address);
-        popupMessageHandler.broadcast(EventType.AccountChanged, address);
+        popupMessageHandler.broadcast(EventType.AccountChanged, address, {
+            nonWhitelistedTabCallback: (tab: chrome.tabs.Tab) => {
+                if (tab.url) {
+                    popupMessageHandler.broadcastToUrl(EventType.AccountDisconnected, tab.url);
```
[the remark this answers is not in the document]
soerenbf: Yeah.. so on second thought, this is not a good solution. Let's discuss this when you have the time (and also @abizjak's comment above)

packages/browser-wallet/CHANGELOG.md
soerenbf: We discussed this, and decided that this is fine as implemented now.

packages/browser-wallet/src/popup/store/account.ts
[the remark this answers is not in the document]
soerenbf: Changed this to:

- Removed disconnect emit from account change if not whitelisted
- Added emission of account change/disconnect to network change handler (if an account on the new network is connected, emit account change, otherwise account disconnect for the corresponding tab)

packages/browser-wallet/src/popup/store/settings.ts
```diff
@@ -27,15 +27,49 @@ const storedNetworkConfigurationAtom = atomWithChromeStorage<NetworkConfiguratio
...
+
+        popupMessageHandler.broadcast(EventType.ChainChanged, networkConfiguration.genesisHash, {
+            requireWhitelist: false,
+            nonWhitelistedTabCallback: (tab) => {
+                if (!tab.url) {
+                    return;
+                }
+
+                // If tab has any account connected, send account changed event, otherwise account disconnected.
+                const [firstConnectedAccount] = sortedConnectedSites?.[0] ?? [];
```
shjortConcordium, line 65: I'm not sure I understand how this checks if the tab is connected, does it not just check whether there are any site connected?

shjortConcordium (changes requested):
I'm not sure what it is, but the first time you do recovery, it now fails because `genesisHash` is accessed before network is chosen:
`Indexed storage should not be accessed before setting the network`

packages/browser-wallet/src/popup/store/settings.ts
soerenbf: Fixed this now.

shjortConcordium: It looks like it should work, but if I switch between mainnet and testnet, I never get "account disconnected" :thinking:

soerenbf: Yeah sorry.. Must have not tested properly here. Should work now.

PR #266, opened 2023-03-28
Fix opening portals causing blank screen 

shjortConcordium:
## Purpose

 - Fix crashes when opening token raw Metadata (and other modals) 
 - Fix bad parameters from dApps crashing the wallet (the request is now rejected instead)

## Changes

Use updated wallet-common-helpers that support both commonJS (for tests) and ES Modules (for the wallet itself)
 - I recently changed that library to use commonJS modules instead of ES, to make the tests run, but it turns out that references to react-dom then stopped working. Now it uses compiles both versions and we use ES modules in the wallet build and commonJS when running the tests now.
 
 To fix the bad parameters crashing the wallet, I've added a check in the background script that rejects the transaction, if we are unable to serialize the parameters.
 - I had to move some things around, because the parse function we use for serialized values between the dApp/background/popup was located in the API, which uses the common-sdk (to avoid injecting wasm on websites), but that causes the background script to crash. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/two-step-transfer/index.html
+++ b/examples/two-step-transfer/index.html
@@ -123,7 +123,7 @@
                                 maxContractExecutionEnergy: 30000n,
                             },
                             {
-                                RequestTransfer: ['1000', '1', '3Y1RLgi5pW3x96xZ7CiDiKsTL9huU92qn6mfxpebwmtkeku8ry'],
+                                RequestTransfer: [1000, '1', '3Y1RLgi5pW3x96xZ7CiDiKsTL9huU92qn6mfxpebwmtkeku8ry'],
                             },
                             getSchema(
                                 twoStepTransferSchema,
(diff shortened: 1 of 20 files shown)


examples/two-step-transfer/index.html
```diff
@@ -123,7 +123,7 @@
                                 maxContractExecutionEnergy: 30000n,
                             },
                             {
-                                RequestTransfer: ['1000', '1', '3Y1RLgi5pW3x96xZ7CiDiKsTL9huU92qn6mfxpebwmtkeku8ry'],
+                                RequestTransfer: [1000, '1', '3Y1RLgi5pW3x96xZ7CiDiKsTL9huU92qn6mfxpebwmtkeku8ry'],
```
shjortConcordium, line 126: Turns out this gave incorrect parameters

packages/browser-wallet-api/src/util.ts
```diff
@@ -35,59 +27,35 @@ function isDataBlob(cand: any): cand is DataBlob {
...
 }
 
+/**
+ * Stringify that acts as an inverse for parse from browser-wallet @shared/utils/payload-helpers
+ */
 export function stringify(input: any) {
     return JSON.stringify(input, replacer);
 }
-
-export function parse(input: string | undefined) {
```
shjortConcordium, line 68: Had to move it, so that it could be used in the background script

soerenbf (approved):
Works as advertised, looks good 😄

PR #267, opened 2023-03-31
Merge 0.9.9 - 0.9.11 into main

shjortConcordium:
## Purpose

Merge the changes from 0.9.8 to 0.9.11 into main.

## Changes

-   Updated text when signing arbitrary data with signMessage.

### Fixed

-   Fixed wallet crashing if sendTransaction contained parameters that could not be serialized. (The request is rejected instead)
-   wallet no longer crashes when opening modals (Like token raw metadata)
-   fixed duplicate key error when sending multiple transactions.
-   fixed crash when going back in transfer flow. (introduced in 0.9.9)
-   Bug for transferring tokens where validation was hardcoded to assume 6 decimals.
-   Bug for transferring tokens where tokens with integer value above 64 bits caused a crash.
-   Added support for signing arbitrary data with signMessage.

## Checklist

      hard-to-understand areas.


packages/browser-wallet-api-helpers/README.md
```diff
@@ -142,7 +142,12 @@ It is possible to sign arbitrary messages using the keys for an account stored i
 
 If the wallet is locked, or you have not connected with the wallet (or previously been whitelisted) or if the user rejects signing the meesage, the `Promise` will reject.
 
-The following exemplifies requesting a signature of a message:
+The message should either utf8 string or an object with the following fields:
```
orhoj, line 145: A word is missing here.
```
The message should contain either a UTF-8 string or an object...
```

packages/browser-wallet-api-helpers/README.md
```diff
@@ -142,7 +142,12 @@ It is possible to sign arbitrary messages using the keys for an account stored i
...
+
+-   data: A hex string representing the bytes that should be signed.
+-   schema: A base64 string that represents a schema for the data field, and which can be used to deserialize the data into a JSON format.
```
orhoj, line 148: What kind of schema? Can we link to somewhere with some help for this? I wouldn't know where to start if I wanted to use this schema functionality. Is it the smart contract schema stuff?

shjortConcordium: I've added "(For reference, the type of schema used here is the same that is used for smart contract types)"
But I'm not sure where would we link them, but this feature is used for sponsored transactions, so we should have an explanation for this.

abizjak: https://developer.concordium.software/en/mainnet/smart-contracts/general/contract-schema.html

PR #268, opened 2023-04-03
Improve CIS-2 lookup errors

shjortConcordium:
## Purpose

Improve error messages when looking up CIS-2 tokens.

## Changes

- Reworked the structure of the `manageTokens` flow, including moving it under the `Tokens` folder.
- Improve the validation function for the `chooseContract` component, to improve the error messages shown.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/pages/Account/i18n/en.ts
```diff
@@ -74,9 +74,13 @@ const t = {
         add: {
             lookupTokens: 'Look for tokens',
             indexRequired: 'Contract index is required',
+            negativeIndex: 'Contract index may not be negative',
```
orhoj, line 77: May not be makes it sounds like a suggestion. Perhaps it should be more firm like:
```
Contract index must be non-negative
```
or something along those lines. Perhaps Danielle has a better phrasing here.

packages/browser-wallet/src/popup/pages/Account/i18n/en.ts
```diff
@@ -74,9 +74,13 @@ const t = {
...
+            invalidIndex: 'Contract index must be a integer',
+            indexMax: 'Contract index can not exceed 18446744073709551615',
             noContractFound: 'No contract found on index',
             noTokensError: 'No tokens found in contract',
-            failedTokensError: 'Error occurred when checking for tokens in the contract',
+            failedTokensError:
+                'The following errors was encountered while checking for tokens in the contract: \n{{ errors }}',
```
orhoj, line 83:
```suggestion
                'The following errors were encountered while checking for tokens in the contract: \n{{ errors }}',
```

packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -178,7 +178,12 @@ export async function getTokenMetadata({ url, hash: checksumHash }: MetadataUrl)
         throw new Error('Metadata does not match checksum provided with url');
     }
 
-    const metadata = JSON.parse(body.toString());
+    let metadata;
+    try {
+        metadata = JSON.parse(body.toString());
+    } catch (e) {
+        throw new Error('Metadata url did not return valid JSON.');
```
orhoj, line 185: Will this error message be shown to the user? If so, then it should have a translation.

packages/browser-wallet/src/popup/pages/Account/i18n/en.ts
shjortConcordium: I'd think we want to avoid non-negative :thinking: 
I've changed it to "...index cannot be negative" to make it sound less suggestive.

orhoj (approved):

PR #271, opened 2023-04-17
Support eID document types

orhoj:
## Purpose
Support identities with generic document types, as these will be received when using Criipto through DTS.

## Changes
- Fixed issues with missing translations for gender, document type and dates on an identity.
- Parsing of identities with a generic document type.
- Better display of long document numbers.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/shared/utils/identity-helpers.ts
```diff
@@ -8,3 +9,71 @@ export function useGetAttributeName() {
...
+            return t('documentType.DriversLicense');
+        case DocumentType.ImmigrationCard:
+            return t('documentType.ImmigrationCard');
+        default:
+            return value;
+    }
+}
+
+export function useDisplayAttributeValue() {
+    const { t, i18n } = useTranslation('shared', { keyPrefix: 'idAttributeValues' });
```
shjortConcordium, line 63: You don't need to make this a hook, you can access the text like this :thinking:
```
import i18n from '@popup/shell/i18n';
...
    return i18n.t('idAttributeValues.documentType.NA');
```

orhoj: If we do that, then we have to check whether translations are actually loaded. See https://github.com/i18next/react-i18next/issues/909. So I believe the hook way is probably simpler.

shjortConcordium (approved):

PR #281, opened 2023-04-19
SendTransaction check for insufficient funds + translate errors

shjortConcordium:
## Purpose

Ensure that `SendTransaction` does not allow transaction to be sent, when they would fail due to insufficient funds.

## Changes

- Check available funds before sending transactions in `sendTransaction`.
- Translate errors in `SendTransaction`.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/shared/utils/transaction-helpers.ts
```diff
@@ -204,3 +208,22 @@ export const createPendingTransactionFromAccountTransaction = (
 export function useHasPendingTransaction(transactionType: AccountTransactionType): boolean {
     return useAtomValue(selectedPendingTransactionsAtom).some((t) => t.type === transactionType);
 }
+
+/**
+ * Extract the microCCD amount related for the transaction, excluding the cost.
+ * Note that for many transactions there is no related amount, in which case this returns 0.
+ */
+export function getTransactionAmount(type: AccountTransactionType, payload: AccountTransactionPayload): bigint {
```
orhoj, line 216: I suppose scheduled transfer and transfer to encrypted are not here because we don't support them in the first place?

orhoj (commented):
I am still able to make transactions that fail, e.g. if I try to wrap all my CCD's.

packages/browser-wallet/src/popup/shared/utils/transaction-helpers.ts
shjortConcordium: Yes, but mostly because their payload types have not been added to the SDK.

shjortConcordium:
> 

This was caused by an off-by-one error in the cost calculation.

orhoj (approved):

PR #287, opened 2023-04-20
Fix behaviour when attempting to creating existing account

shjortConcordium:
## Purpose

Fix bad behaviour, which has recently worsened, when creating an existing account.

## Changes

- Add existing account when trying to recreate it.
- Add missing await in credential-deployment.
- Fix `monitorCredentialStatus` not halting when "transaction not found".

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/package.json
+++ b/packages/browser-wallet/package.json
@@ -18,6 +18,7 @@
         "@concordium/browser-wallet-api": "workspace:^",
         "@concordium/browser-wallet-api-helpers": "workspace:^",
         "@concordium/web-sdk": "^3.4.0",
+        "@protobuf-ts/runtime-rpc": "^2.8.2",
         "@scure/bip39": "^1.1.0",
         "axios": "^0.27.2",
         "buffer": "^6.0.3",
--- a/packages/browser-wallet/src/background/confirmation.ts
+++ b/packages/browser-wallet/src/background/confirmation.ts
@@ -51,20 +51,6 @@ async function monitorCredentialStatus(initialNetwork: NetworkConfiguration, cre
 
         try {
             const response = await client.getBlockItemStatus(deploymentHash);
-            // transaction has been discarded by the node.
-            if (!response) {
-                await updateCredentials(
-                    [
-                        {
-                            ...info,
-                            status: CreationStatus.Rejected,
-                        },
-                    ],
-                    genesisHash
-                );
-                return false;
-            }
-
             // transaction has not finalized yet
             if (response?.status !== TransactionStatusEnum.Finalized) {
                 return true;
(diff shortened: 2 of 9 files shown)


packages/browser-wallet/src/popup/pages/AddAccount/Confirm.tsx
```diff
@@ -94,6 +98,12 @@ export default function Confirm() {
     // TODO: Better faking of AccountDetails
     return (
         <div className="flex-column align-center">
+            <Modal open={Boolean(accountRecovered)} disableClose>
+                <p>{t('accountRecovered')}</p>
+                <Button width="wide" onClick={() => nav(absoluteRoutes.home.account.path)}>
```
orhoj, line 103: This navigates to the account overview and not the specific account.

orhoj (changes requested):

packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -80,7 +66,20 @@ async function monitorCredentialStatus(initialNetwork: NetworkConfiguration, cre
                 genesisHash
             );
             return false;
-        } catch {
+        } catch (e) {
+            // transaction has been discarded by the node.
+            if ((e as Error).message.includes('transaction%20not%20found')) {
```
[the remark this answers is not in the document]
abizjak: The node returns a proper status code https://docs.rs/tonic/latest/tonic/struct.Status.html#method.not_found when it cannot find the transaction. Shouldn't that be used here?

packages/browser-wallet/package.json
```diff
@@ -18,6 +18,7 @@
         "@concordium/browser-wallet-api": "workspace:^",
         "@concordium/browser-wallet-api-helpers": "workspace:^",
         "@concordium/web-sdk": "^3.4.0",
+        "@protobuf-ts/runtime-rpc": "^2.8.2",
```
orhoj, line 21: Latest version is `2.8.3`. Any particular reason for not using that over this?

orhoj (approved):

packages/browser-wallet/src/background/confirmation.ts
```diff
@@ -80,7 +67,20 @@ async function monitorCredentialStatus(initialNetwork: NetworkConfiguration, cre
                 genesisHash
             );
             return false;
-        } catch {
+        } catch (e) {
+            // transaction has been discarded by the node.
+            if ((e as RpcError).code === 'NOT_FOUND') {
```
orhoj, line 72: It would be preferable if the `NOT_FOUND` was gotten directly from an enum in the library, but as far as I can see it is not available - they just link to the grpc repository file with the enum.

PR #304, opened 2023-05-01
Ensure when validating baker/delegation stake reduction that atDisposal can cover fee

shjortConcordium:
## Purpose

Fix when reducing baker/delegation stake, that a user could continue and then get rejected by the node.

## Changes

Validate that current atDisposal can cover the fee.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/shared/utils/transaction-helpers.ts
+++ b/packages/browser-wallet/src/popup/shared/utils/transaction-helpers.ts
@@ -23,6 +23,7 @@ import {
     displayAsCcd,
     fractionalToInteger,
     isValidCcdString,
+    getPublicAccountAmounts,
 } from 'wallet-common-helpers';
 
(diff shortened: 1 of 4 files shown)


packages/browser-wallet/src/popup/shared/utils/transaction-helpers.ts
```diff
@@ -133,6 +140,11 @@ export function validateDelegationAmount(
         return i18n.t('utils.ccdAmount.insufficient');
     }
 
+    // the fee must be paid with the current funds at disposal, because a reduction in delegation amount is not immediate.
+    if (getPublicAccountAmounts(accountInfo).atDisposal < estimatedFee) {
+        return i18n.t('utils.ccdAmount.insufficientFundsAtDisposal');
```
orhoj, line 145: The error message should be the same whenever the user has insufficient funds.

PR #308, opened 2023-05-08
Wallet api types/lax schematype

soerenbf:
This makes it possible to construct an object adhering to the `SchemaWithContext` type from the JS SDK, to make it easier for a user to directly use the SDK to call the wallet API.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
+++ b/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
@@ -39,13 +39,14 @@ export enum EventType {
     ChainChanged = 'chainChanged',
 }
 
+export type SchemaTypeString = 'module' | 'parameter';
 export enum SchemaType {
     Module = 'module',
     Parameter = 'parameter',
 }
 
 export type SchemaWithContext = {
-    type: SchemaType;
+    type: SchemaType | SchemaTypeString;
     value: string;
 };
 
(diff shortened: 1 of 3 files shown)


packages/browser-wallet-api-helpers/src/wallet-api-types.ts
```diff
@@ -39,13 +39,14 @@ export enum EventType {
     ChainChanged = 'chainChanged',
 }
 
+export type SchemaTypeString = 'module' | 'parameter';
 export enum SchemaType {
     Module = 'module',
     Parameter = 'parameter',
 }
 
 export type SchemaWithContext = {
```
shjortConcordium, line 48:
```suggestion

export enum SchemaType {
    Module = 'module',
    Parameter = 'parameter',
}

type LaxStringEnumValue<E extends string> = `${E}`;

export type SchemaWithContext = {
    type: LaxStringEnumValue<SchemaType>;
    value: string;
};
```
then SchemaWithContext.type would be `'module' | 'parameter'` but without duplicating the keys.

soerenbf: Hmm.. didn't think this would be possible without casting f.x. `SchemaType.Parameter` to a string, but it seems it works.. guess I remember wrong!

shjortConcordium: Sorry, but I couldn't do a suggestion that would actually work.

soerenbf: For once, I trusted the suggestion.. never again!

Should be fixed now though.

soerenbf:
@shjortConcordium anything happening here?

shjortConcordium (approved):
I guess I forgot about it. :shrug:

PR #316, opened 2023-05-25
Fix grpc client first call failing

shjortConcordium:
## Purpose

Before, the URL was not actually immediately used, instead the client would always use the previous call's URL, which on the first call is undefined. It also meant that after changing network, the first call would still be directed to the old network. 
This should fix that.

## Changes

Apply base settings immediately before performing the actual call.

## Checklist

      hard-to-understand areas.


orhoj:
I still experience the bug regarding changing networks.

orhoj (approved):

PR #319, opened 2023-06-19
Add mainnet to wCCD example

DOBEN:
## Purpose

Add support for wCCD frontend for mainnet. wCCD is now deployed on mainnet and we would like to host this frontend for it as well.

## Changes

The front end can either be built for mainnet or testnet as described in READ.me. Equally, the `dockerFile` can be built for mainnet or testnet.

- If the user uses the browser wallet, an error message will appear now if the browser wallet is set to the wrong network.
- If the user uses the mobile wallet (Android), an error message will appear now if the wrong (for the wrong network) mobile wallet app is used.


examples/wCCD/src/constants_mainnet_network.ts
```diff
@@ -0,0 +1,20 @@
+// The MAINNEt_GENESIS_BLOCK_HASH is used to check that the user has its browser wallet connected to testnet and not to mainnet.
+import { Network } from '@concordium/react-components';
+
+export const MAINET_GENESIS_BLOCK_HASH = '9dd9ca4d19e9393877d2c44b70f89acbfc0883c2243e5eeaecc0d1cd0503f478';
```
abizjak, line 4: It's usually spelled "MAINNET".

abizjak (commented):
I think the Jenkinsfile also must be updated to take into account  the new build parameter.

examples/wCCD/src/constants_network.ts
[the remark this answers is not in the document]
DOBEN: I tried that but run into the issue that building of the frontend works but once run in the browser wallet I get this error `process` not defined. Is there a specific environmental variable framework that needs to be used to get this working?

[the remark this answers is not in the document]
shjortConcordium: The `process.env.ENV_VAR_NAME` only "exists" during "compilation", so you have to replace it in the build script.
Simplest way is in esbuild.config.ts to add:
```
const config: BuildOptions = {
    ...
    define: {
        ...
        'process.env.network': process.env.network
    },
}
```
So access the value during in the app.

examples/wCCD/src/wCCD.tsx
```diff
@@ -133,10 +133,23 @@ export default function wCCD(props: WalletConnectionProps) {
     const [admin, setAdmin] = useState<string>();
     const [adminError, setAdminError] = useState('');
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const testnet = 'testnet';
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const mainnet = 'mainnet';
```
shjortConcordium, line 139:
```suggestion
```
Or why are you keeping them? :thinking:

shjortConcordium (approved):

abizjak (approved):

examples/wCCD/src/wCCD.tsx
DOBEN: `process.env.NETWORK` is replaced by e.g. testnet (not by the string `testnet`) and without defining testnet=`testnet`, I got an error that testnet variable is not defined.

shjortConcordium (approved):

PR #324, opened 2023-07-07
browser-wallet-api-helpers: Make dependency to web-sdk a peer dependency

bisgardo:
*This is a proposal that hasn't been discussed yet. There may be reasons for not doing this that I'm not aware of.*

## Purpose

Having the dependency be "standard" means that the SDK version can be bumped only via this library. Otherwise, when the consuming app adds an explicit dependency to the SDK, it will include both versions which may clash and otherwise behave in unexpected ways.

## Changes

Make `browser-wallet-api-helpers` dependency to web-sdk a peer dependency.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api-helpers/CHANGELOG.md
+++ b/packages/browser-wallet-api-helpers/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 3.0.0
+
+### Changed
+
+-   Make dependency to `@concordium/web-sdk` a peer dependency.
+    This will require consuming apps to add an explicit dependency to the library with its own version constraints.
+
 ## 2.5.1
 
 ### Changed
--- a/packages/browser-wallet-api-helpers/package.json
+++ b/packages/browser-wallet-api-helpers/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@concordium/browser-wallet-api-helpers",
-    "version": "2.5.1",
+    "version": "3.0.0",
     "license": "Apache-2.0",
     "packageManager": "yarn@3.2.0",
     "main": "lib/index.js",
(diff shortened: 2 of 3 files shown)


soerenbf (commented):
I think it makes sense, however I would also like to hear the general opinion of the other reviewers before approving 😄

bisgardo:
Ping @shjortConcordium @orhoj. I'll fix the build errors if you agree that we should move forward with this. If we don't do this, then the only way to override the dependency is to set a yarn resolution. Which I think is something you should try to avoid.

shjortConcordium (commented):
The web-sdk is only used for the types here, so the "unexpected" behaviour would be limited to type errors.
The argument for not changing it to peerDependency is that the injected code from the wallet is (should be?) the same version as the helpers package.

If we get annoying type errors, we can fix those, I think, like the transaction types issue, I relax the type in #327.

Thought @orhoj :thinking:

bisgardo:
> The web-sdk is only used for the types here, so the "unexpected" behaviour would be limited to type errors.

That would be the same if the dependency was set by the client?

> The argument for not changing it to peerDependency is that the injected code from the wallet is (should be?) the same version as the helpers package.

What code is injected by the wallet? What guarantees does the explicit dependency give us about that? Don't they run in completely separate execution environments?

shjortConcordium:
> That would be the same if the dependency was set by the client?

I guess, I was commenting on the reasoning stated in the issue, that this would prevent "otherwise behave in unexpected ways".

> What code is injected by the wallet? What guarantees does the explicit dependency give us about that? Don't they run in completely separate execution environments?

The walletApi, which the package has types for, is injected into the website's env by the wallet :thinking:

soerenbf:
Where did we land with this? We're releasing the next major version of the web SDK soon, maybe it could make sense to revisit this proposal?

soerenbf:
@bisgardo @shjortConcordium @orhoj 

FYI the web SDK is made a peer dependency in #395 as well. I ran into some issues with type mismatches. It makes sense to define it as a peer dependency, as parts of the API require users to use the web SDK (f.x. specify transaction type when sending transactions).

I think we should close this PR.

bisgardo:
Covered by https://github.com/Concordium/concordium-browser-wallet/pull/395.

PR #327, opened 2023-07-20
Relax AccountTransactionType in wallet-api

shjortConcordium:
## Purpose

Relax transaction type parameter in the `wallet-api`, so numbers can be given directly and tsc doesn't complain if a different version of the enum is used.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
+++ b/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
@@ -10,6 +10,7 @@ import type {
     IdProofOutput,
     ConcordiumGRPCClient,
 } from '@concordium/web-sdk';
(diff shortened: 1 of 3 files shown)


packages/browser-wallet-api-helpers/src/wallet-api-types.ts
```diff
@@ -45,6 +45,7 @@ export enum SchemaType {
 }
 
 type LaxStringEnumValue<E extends string> = `${E}`;
+type LaxNumberEnumValue<E extends number> = `${E}` extends `${infer T extends number}` ? T : never;
```
orhoj, line 48: Please unit test this.

orhoj (changes requested):

orhoj (approved):

PR #338, opened 2023-08-16
Extra details

orhoj:
## Purpose
Add a button for showing "extra details" about a credential, i.e. non-attribute information.

## Changes
- Added details button to more menu for credentials in the details view.
- Adjusted popup menu component to support what's needed.

## Checklist
      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/VerifiableCredential/VerifiableCredentialDetails.tsx
+++ b/packages/browser-wallet/src/popup/pages/VerifiableCredential/VerifiableCredentialDetails.tsx
@@ -152,25 +152,33 @@ export default function VerifiableCredentialDetails({
             return undefined;
         }
 
-        const detailsButton = {
-            title: t('menu.details'),
-            onClick: () => setShowExtraDetails(true),
-        };
+        const menuButtons = [];
 
-        let revokeButton;
         if (credentialEntry?.credentialInfo.holderRevocable && status !== VerifiableCredentialStatus.Revoked) {
-            revokeButton = {
+            const revokeButton = {
                 title: t('menu.revoke'),
                 icon: <RevokeIcon />,
                 onClick: goToConfirmPage,
             };
+            menuButtons.push(revokeButton);
         }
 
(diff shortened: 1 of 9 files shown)


packages/browser-wallet/src/popup/pages/VerifiableCredential/i18n/da.ts
```diff
@@ -7,6 +7,12 @@ const t: typeof en = {
     },
     menu: {
         revoke: 'Ophæv',
+        details: 'Detaljer',
+    },
+    details: {
+        id: 'Legimitationholders ID',
```
shjortConcordium, line 13:
```suggestion
        id: 'Legitimationholders ID',
```

packages/browser-wallet/src/popup/pages/VerifiableCredential/VerifiableCredentialDetails.tsx
```diff
@@ -121,18 +182,31 @@ export default function VerifiableCredentialDetails({
         <>
             <Topbar
                 title={t('topbar.details')}
-                backButton={{ show: true, onClick: backButtonOnClick }}
+                backButton={{
+                    show: true,
+                    onClick: () => (showExtraDetails ? setShowExtraDetails(false) : backButtonOnClick()),
+                }}
                 menuButton={menuButton}
```
shjortConcordium, line 189: I can still click on details, when I'm on the details page. :thinking: 
I think either that should be removed or changed to be a "see card front side" button. (If we change the button, maybe the back button should be changed to not treat details as a subpage)

shjortConcordium (approved):

packages/browser-wallet/src/popup/pages/VerifiableCredential/VerifiableCredentialDetails.tsx
shjortConcordium: I don't like it but I'll have to go along with it

PR #339, opened 2023-08-16
Change from json-rpc to grpc

DOBEN:
## Purpose

- JSON-RPC is deprecated and will be eventually removed from all our libraries. This PR migrates the eSealing dApp.

## Changes

- Replace using jsonRPC with gRPC

### DIFF
--- a/examples/eSealing/src/eSealing.tsx
+++ b/examples/eSealing/src/eSealing.tsx
@@ -156,7 +156,7 @@ export default function SEALING(props: WalletConnectionProps) {
 
     useEffect(() => {
         // View file record.
-        if (grpcClient && account && fileHashHex !== '') {
+        if (grpcClient && fileHashHex !== '') {
             viewFile(grpcClient, fileHashHex)
                 .then((record) => {
                     setGetFileError('');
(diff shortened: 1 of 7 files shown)


examples/eSealing/src/utils.ts
```diff
@@ -25,10 +25,10 @@ export async function register(
             receiveName: `${E_SEALING_CONTRACT_NAME}.registerFile`,
             maxContractExecutionEnergy: 30000n,
         } as UpdateContractPayload,
```
bisgardo, line 27:
```suggestion
        },
```

examples/eSealing/src/eSealing.tsx
```diff
@@ -155,8 +156,8 @@ export default function SEALING(props: WalletConnectionProps) {
 
     useEffect(() => {
         // View file record.
-        if (connection && account && fileHashHex !== '') {
-            withJsonRpcClient(connection, (rpcClient) => viewFile(rpcClient, fileHashHex))
+        if (grpcClient && account && fileHashHex !== '') {
```
bisgardo, line 159: I know this it isn't part of this change, but should `account` be included here (and in the dependency list)? It's not being used inside.

bisgardo (approved):
LGTM

examples/eSealing/src/eSealing.tsx
DOBEN: I removed the `account` as a dependency. It is not needed here.

PR #340, opened 2023-08-16
Replace json-rpc with grpc in voting dApp

DOBEN:
## Purpose
- JSON-RPC is deprecated and will be eventually removed from all our libraries. This PR migrates the voting dApp.

## Changes
-  Replace using jsonRPC with gRPC

### DIFF
--- a/examples/voting/package.json
+++ b/examples/voting/package.json
@@ -1,11 +1,11 @@
 {
     "name": "voting",
     "license": "Apache-2.0",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "packageManager": "yarn@3.2.0",
     "dependencies": {
         "@concordium/browser-wallet-api-helpers": "^2.0.0",
-        "@concordium/web-sdk": "^3.3.1",
+        "@concordium/web-sdk": "^6.0.0",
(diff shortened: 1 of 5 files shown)


examples/voting/package.json
```diff
@@ -1,11 +1,11 @@
 {
     "name": "voting",
     "license": "Apache-2.0",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "packageManager": "yarn@3.2.0",
     "dependencies": {
         "@concordium/browser-wallet-api-helpers": "^2.0.0",
-        "@concordium/web-sdk": "^3.3.1",
+        "@concordium/web-sdk": "^6.0.0",
```
bisgardo, line 8: I think you "need" to do this as a resolution. Otherwise you'll get 2 SDK versions in play at the same time; this one and [another one](https://github.com/Concordium/concordium-browser-wallet/blob/main/packages/browser-wallet-api-helpers/package.json#L22) from `browser-wallet-api-helpers`. It works most of the time but could cause problems down the line.

bisgardo (approved):

abizjak (approved):

PR #348, opened 2023-08-23
Display pending verifiable credentials

shjortConcordium:
## Purpose

Display pending verifiable credentials.

## Changes

- Add session storage for verifiable credential metadata.
- Use session stored metadata if credentialEntry is missing.
- Catch some errors in VC hooks.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/AddWeb3IdCredential/AddWeb3IdCredential.tsx
+++ b/packages/browser-wallet/src/popup/pages/AddWeb3IdCredential/AddWeb3IdCredential.tsx
@@ -6,7 +6,7 @@ import { useLocation } from 'react-router-dom';
 import ExternalRequestLayout from '@popup/page-layouts/ExternalRequestLayout';
 import Button from '@popup/shared/Button';
 import {
-    sessionTemporaryVerifiableCredentialMetadataAtom,
+    sessionTemporaryVerifiableCredentialMetadataUrlsAtom,
     sessionTemporaryVerifiableCredentialsAtom,
     storedVerifiableCredentialMetadataAtom,
     storedVerifiableCredentialsAtom,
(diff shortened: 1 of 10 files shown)


orhoj:
If I press the pending credential to see its details, then I get shown blank content.

packages/browser-wallet/src/popup/pages/AddWeb3IdCredential/AddWeb3IdCredential.tsx
```diff
@@ -144,18 +146,20 @@ export default function AddWeb3IdCredential({ onAllow, onReject }: Props) {
...
-            id: createCredentialId(credentialHolderId, issuer, network),
+            id: credentialId,
             index,
         };
         await setWeb3IdCredentials([...web3IdCredentials.value, fullCredential]);
         if (metadata) {
             const newMetadata = { ...verifiableCredentialMetadata.value };
             newMetadata[metadataUrl.url] = metadata;
             await setVerifiableCredentialMetadata(newMetadata);
+            await setTempMetadata({ ...tempMetdata.value, [credentialId]: metadata });
```
orhoj, line 162: This is probably fine for now, but I think we could avoid duplicating the metadata. 

The problem is that we can't find the `credentialEntry` so the metadata hook could also look somewhere else for the metadata URL, which should be all that is necessary to store to fix problem.

packages/browser-wallet/src/shared/storage/types.ts
```diff
@@ -30,6 +30,7 @@ export enum ChromeStorageKey {
     VerifiableCredentialSchemas = 'verifiableCredentialSchemas',
     VerifiableCredentialMetadata = 'verifiableCredentialMetadata',
     TemporaryVerifiableCredentials = 'tempVerifiableCredentials',
+    TemporaryVerifiableCredentialMetadataUrls = 'tempVerifiableCredentialMetadata',
```
orhoj, line 33:
```suggestion
    TemporaryVerifiableCredentialMetadataUrls = 'tempVerifiableCredentialMetadataUrls',
```

orhoj (approved):

PR #356, opened 2023-08-25
Fix import popup

orhoj:
## Purpose
Fix import flow on Windows and macOS (and improve on the race condition).

## Changes
- Wait for the popup to close before opening the import window.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/background/web3Id.ts
```diff
@@ -194,7 +194,35 @@ export const runIfValidWeb3IdProof: RunCondition<MessageStatusWrapper<undefined>
...
+                }
+                escapeCounter += 1;
+                setTimeout(waitForClosed, 100);
+            }
+        }, 100);
+    });
+}
+
 async function loadWeb3IdBackup(): Promise<void> {
+    await waitForClosedPopup(30);
```
shjortConcordium, line 225: 30? could we maybe move this to a constant instead of defining magic constants inside the code :shrimp:

shjortConcordium (approved):

PR #374, opened 2023-09-01
Use the schema to validate a credential being added

orhoj:
## Purpose
Actually use the schema to validate a credential being added. This will e.g. block giving a string for an integer attribute.

## Changes
- Validate credential with the JSON schema.

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/pages/AddWeb3IdCredential/AddWeb3IdCredential.tsx
```diff
@@ -115,23 +116,31 @@ export default function AddWeb3IdCredential({ onAllow, onReject }: Props) {
...
-            const requiredAttributes = schema.properties.credentialSubject.properties.attributes.required;
-            for (const requiredAttribute of requiredAttributes) {
-                if (!Object.keys(credential.credentialSubject.attributes).includes(requiredAttribute)) {
-                    missingRequiredAttributeKeys.push(requiredAttribute);
+            // Use the schema to validate the credential.
+            const validator = new Validator();
+            try {
+                const validationResult = validator.validate(
+                    { credentialSubject: credential.credentialSubject },
+                    schema as unknown as Schema
```
shjortConcordium, line 124: This does allow me to do timestamp attributes as dates. Did we want allow that here or only in statements?

orhoj: I am not sure.

orhoj: Ah, I understand now. It does not allow it now. Also if I add the object type, then it is not displayed correctly (shows as [object Object].

orhoj: Going to as the core team about it. We will leave this out until we know.

orhoj: @shjortConcordium we decide how we want it. So I would be interested in your opinion. Restricting to the schema when adding would also fix the serialization issue we had (and already fixed).

shjortConcordium (approved):

PR #385, opened 2023-09-05
Allow bigints as type integer

orhoj:
## Purpose
Fix issue where credentials with integer values in their schema could not be added.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/background/web3Id.ts
+++ b/packages/browser-wallet/src/background/web3Id.ts
@@ -134,12 +134,15 @@ function timestampToDate(attribute: TimestampAttribute): Date {
 }
 
 function validateTimestampAttribute(attributeTag: string, attributeValue: AttributeType) {
-    if (
-        isTimestampAttribute(attributeValue) &&
-        (timestampToDate(attributeValue).getTime() < MIN_DATE_TIMESTAMP ||
-            timestampToDate(attributeValue).getTime() > MAX_DATE_TIMESTAMP)
-    ) {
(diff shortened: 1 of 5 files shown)


packages/browser-wallet/src/background/web3Id.ts
```diff
@@ -126,12 +128,21 @@ function rejectRequest(message: string): { run: false; response: MessageStatusWr
...
-    if (
-        attributeValue instanceof Date &&
-        (attributeValue.getTime() < MIN_DATE_TIMESTAMP || attributeValue.getTime() > MAX_DATE_TIMESTAMP)
-    ) {
-        return `The attribute [${attributeValue}] for key [${attributeTag}] is out of bounds for a Date. The Date must be between ${MIN_DATE_ISO} and ${MAX_DATE_ISO}`;
+    if (isTimestampAttribute(attributeValue)) {
+        const timestamp = timestampToDate(attributeValue).getTime();
+        if (Number.isNaN(timestamp)) {
+            return `The attribute [${attributeValue.timestamp}] for key [${attributeTag}] cannot be parsed as a Date.`;
+        }
```
shjortConcordium, line 141: Looks good, only issue i can find is that this allows for Non-standard parsing. (like date "1000")
But we can probably live with that.

shjortConcordium (approved):

packages/browser-wallet/src/background/web3Id.ts
orhoj: I agree we can live with that.

PR #393, opened 2023-10-06
Replace jsonRPC with gRPC

DOBEN:
## Purpose

Replace jsonRPC with gRPC in wCCD dApp

## Changes

Replace jsonRPC with gRPC in wCCD dApp


examples/wCCD/package.json
```diff
@@ -1,10 +1,10 @@
 {
     "name": "wccd",
     "packageManager": "yarn@3.2.0",
-    "version": "4.0.0",
+    "version": "4.1.0",
     "license": "Apache-2.0",
     "dependencies": {
-        "@concordium/react-components": "^0.2.0",
+        "@concordium/react-components": "^0.3.0",
```
bisgardo, line 7: The incompatibility with the Android wallet happened because this line caused `@concordium/wallet-connectors:0.3.1` to be used which sends the schema in a broken format.

The Android wallet turns out to not implement that format correctly (fixed in https://github.com/Concordium/concordium-reference-wallet-android/pull/304). New projects should pick `wallet-connectors:0.3.2` but it seems like the broken version was already in `yarn.lock` or something which is why that didn't happen (`react-components:0.3.0` depends on `^0.3.1` of `wallet-connectors` so will stick with any existing version satisfying that).

I added a commit to bump the lib in the lock file and now it works in the current version of the Android wallet.

DOBEN: great thanks for that.

soerenbf (approved):
Tested wrap/unwrap, and they still work - LGTM 😄

examples/wCCD/src/wCCD.tsx
```diff
@@ -138,25 +146,29 @@ export default function wCCD(props: WalletConnectionProps) {
     const mainnet = 'mainnet';
     let WCCD_CONTRACT_INDEX: bigint;
 
+    let grpcClient: ConcordiumGRPCClient | undefined;
+
     if (process.env.NETWORK === mainnet) {
         WCCD_CONTRACT_INDEX = WCCD_CONTRACT_INDEX_MAINNET;
```
bisgardo, line 152: Not a fan of assigning what looks like a constant. Would prefer constructing a configuration object (that also included the `Network` object).

bisgardo (approved):
LGTM

examples/wCCD/src/wCCD.tsx
```diff
@@ -138,25 +146,29 @@ export default function wCCD(props: WalletConnectionProps) {
     const mainnet = 'mainnet';
     let WCCD_CONTRACT_INDEX: bigint;
 
+    let grpcClient: ConcordiumGRPCClient | undefined;
+
     if (process.env.NETWORK === mainnet) {
```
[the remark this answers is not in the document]
DOBEN: Simplified the handling of mainnet/testnet by resolving it at the top level and passing it down to the different components.

PR #394, opened 2023-10-06
Add Sliders for commissionFields

shjortConcordium:
## Purpose

Enable changing commission rates, instead of being forced to use the minimum rate.

## Changes

- Add sliders
- Adjustments to commission rates page.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/shared/Form/Slider/Slider.tsx
+++ b/packages/browser-wallet/src/popup/shared/Form/Slider/Slider.tsx
@@ -15,6 +15,7 @@ interface Props extends CommonFieldProps {
     onBlur?(): void;
     className: string;
     name: string;
+    isInvalid?: boolean;
 }
 
 export default function Slider({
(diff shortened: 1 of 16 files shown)


packages/browser-wallet/src/popup/shared/Form/Slider/Slider.tsx
```diff
@@ -0,0 +1,102 @@
...
+        onChange(innerValue);
+    }, [innerValue, onChange]);
+
+    const handleChange = (v: string | undefined) => {
+        if (v === undefined) {
+            setInnerValue(0);
+            return;
+        }
+
+        const parser = Number.isInteger(step) ? parseInt : parseFloat;
```
orhoj, line 49: There is some weird behaviour where I cannot create a decimal number from scratch.

orhoj (changes requested):

packages/browser-wallet/src/popup/shared/utils/baking-helpers.ts
```diff
@@ -13,10 +13,13 @@ const fractionResolution = 100000;
 const percentageModifier = fractionResolution / 100;
 
 export const decimalToRewardFraction = (d: number | undefined) =>
-    d === undefined ? undefined : d * percentageModifier;
+    d === undefined ? undefined : Math.floor(d * percentageModifier);
```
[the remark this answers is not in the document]
shjortConcordium: I'll revert this. I added it because before I allowed too many decimals, which would cause the wallet to crash. But now I actually ensure that we don't set rates with too many decimals, so this is not necessary.

packages/browser-wallet/src/popup/shared/Form/Slider/Slider.tsx
shjortConcordium: Yes, I tried to emulate the InlineNumber, but that seems to not really work, so I've ported it over instead.

shjortConcordium:
I changed the default to max, and ensured you can't continue with invalid numbers.

packages/browser-wallet/src/popup/shared/Form/Slider/Slider.tsx
shjortConcordium: I've fixed that now.

orhoj (approved):

PR #396, opened 2023-10-23
Support web-sdk v7

limemloh:
## Purpose

This will enable support of newer versions of the web-sdk (v7) without any breaking changes.

## Changes

Replace `CcdAmount` with broader type `HasMicroCcd` which works with both old and new version of web-sdk.

Context: Since v7 of `web-sdk` removes methods on `CcdAmount` it cannot be used with the current released version of the wallet api (helpers), since this uses the old type which require these methods. Since the wallet does not rely on these methods and only on the internal field `microCcdAmount` of `CcdAmount` we can support both versions by reflecting this in the type.

## Checklist

      hard-to-understand areas.


soerenbf (approved):
Looks good :) Could you bump version to 2.8.1 as well and make the release? If you don't have permission to release on NPM, I can also do that.

shjortConcordium:
**** you
(No longer seems relevant)

PR #401, opened 2023-11-06
Tokenomics renaming

shjortConcordium:
## Purpose

Renaming for Tokenomics update.

## Changes

-   Baker/baking renamed to validator/validation in UI.
-   Default name for key file is now `validator-credentials.json`.
-   Hide finalization commission rate in validator flows.

## Checklist

      hard-to-understand areas.


orhoj:
Transaction elements still show as `Configure baker` and `Configure delegation`.

packages/browser-wallet/src/popup/pages/Account/i18n/en.ts
```diff
@@ -94,72 +94,72 @@ const t = {
...
             },
             '4': {
                 title: 'Update pool settings',
-                body: 'There is a cool-down period if you choose to close a pool. During this time, all delegators continue in the pool, and the pool continues to earn rewards if it bakes.\n\nAfter the cool-down, the delegators are removed and the pool will be closed. The baker will keep baking with your own stake afterwards.\n\nChanges to the metadata URL take effect from the next pay day unless the transaction is made too close to the pay day. In that case, it will take effect from the next pay day over.',
+                body: 'There is a cool-down period if you choose to close a pool. During this time, all delegators continue in the pool, and the pool continues to earn rewards if it bakes.\n\nAfter the cool-down, the delegators are removed and the pool will be closed. The validator will keep validation with your own stake afterwards.\n\nChanges to the metadata URL take effect from the next pay day unless the transaction is made too close to the pay day. In that case, it will take effect from the next pay day over.',
```
orhoj, line 139: I don't think you can say "The validator will keep validation with your own stake aftewards".

orhoj:
I can make an empty pool update and then I get this error:

Error: Finalization reward commission rate is not within the allowed range.

orhoj (changes requested):

shjortConcordium:
> Transaction elements still show as `Configure baker` and `Configure delegation`.

I fixed the baker transactions... But `Configure delegation` shouldn't be changed AFAIK.

> I can make an empty pool update (on testnet with singleton intervals) and then I get this error:
> 
> Error: Finalization reward commission rate is not within the allowed range.

Yes, I figured out why it only happened on testnet, I used the incorrect range :upside_down_face:

packages/browser-wallet/src/popup/pages/Account/i18n/en.ts
shjortConcordium: Changed it to "The validator will keep validating with your own stake aftewards".

orhoj (approved):

PR #404, opened 2023-11-07
Remove usage of instanceof when sanitizing wallet-api inputs

shjortConcordium:
## Purpose

Fix potential errors caused by version mismatch in SDK version with dApps.

## Changes

- Remove usage of instanceOf when sanitizing wallet-api inputs.
- Use AccountAddress in piggybank.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api/test/compatibility.test.ts
+++ b/packages/browser-wallet-api/test/compatibility.test.ts
@@ -501,7 +501,7 @@ describe(sanitizeSendTransactionInput, () => {
         };
         const result = sanitizeSendTransactionInput(accountAddress, type, payload);
         expect(result).toEqual(expected);
-        expect(result.payload).toBe(payload);
+        expect((result.payload as RegisterDataPayload).data).toStrictEqual(payload.data);
     });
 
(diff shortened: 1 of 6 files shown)


orhoj (approved):

packages/browser-wallet-api/test/compatibility.test.ts
```diff
@@ -501,7 +503,33 @@ describe(sanitizeSendTransactionInput, () => {
...
+            accountAddress: AccountAddress.fromBase58(accountAddress),
+            type,
+            payload: {
+                data: new DataBlob(Buffer.from('This is data!')),
+            },
+        };
+
+        const result = sanitizeSendTransactionInput(accountAddress, type, payload);
+
+        const parsed = jsonParse(jsonStringify(result));
```
orhoj, line 529: Isn't this overkill for this test? Just doing the sanitize should do the trick?

shjortConcordium: No, because just checking `expect(result).toEqual(expected);`
 doesn't catch the error that I ran into. :upside_down_face: 
 (Also that version of the test succeeds on main)

PR #415, opened 2023-12-06
Make CIS-2 token lookup more lenient

shjortConcordium:
## Purpose

Make CIS-2 token lookup more lenient

## Changes

- Redo `getTokens` so that metadataURL and balances are fetched individually.
- Log errors from token lookup, don't show them in UI.
- Don't fail "Choose Contract" unless all tokens are missing either metadata or balance.
- `fetchTokensConfigure` now keeps looking until 1 valid token is found (/or all tokens in the contract have been checked)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/Account/Tokens/ManageTokens/state.ts
+++ b/packages/browser-wallet/src/popup/pages/Account/Tokens/ManageTokens/state.ts
@@ -65,7 +65,7 @@ export const contractTokensAtom = (() => {
                     set(loadingAtom, true);
 
                     const tokens = get(listAtom);
-                    const topId = [...tokens].reverse()[0]?.pageId;
+                    const topId = tokens[tokens.length - 1]?.pageId;
 
                     const { hasMore, tokens: next } = await fetchTokens(topId);
 
--- a/packages/browser-wallet/src/popup/store/token.ts
+++ b/packages/browser-wallet/src/popup/store/token.ts
@@ -6,6 +6,7 @@ import { ChromeStorageKey, TokenIdAndMetadata, TokenMetadata, TokenStorage } fro
 import { ContractBalances, getContractBalances } from '@shared/utils/token-helpers';
 import { addToastAtom } from '@popup/state';
 import { ContractAddress } from '@concordium/web-sdk';
+import { getContractName } from '@shared/utils/contract-helpers';
 import { AsyncWrapper, atomWithChromeStorage } from './utils';
 import { grpcClientAtom } from './settings';
 import { selectedAccountAtom } from './account';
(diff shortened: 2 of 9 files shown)


orhoj:
The balance shows incorrectly when adding a token (it shows as 0 even if I have tokens).

packages/browser-wallet/src/popup/store/token.ts
```diff
@@ -136,14 +137,22 @@ const cbf = atomFamily<string, Atom<ContractBalances>>((identifier: string) => {
...
             if (tokenIds.length !== 0) {
-                const balances = await getContractBalances(
-                    client,
-                    BigInt(contractIndex),
-                    0n,
-                    tokenIds,
-                    accountAddress,
-                    (error) => set(addToastAtom, error)
-                );
+                const instanceInfo = await client.getInstanceInfo(ContractAddress.create(BigInt(contractIndex), 0n));
```
orhoj, line 140:
```suggestion
                const instanceInfo = await client.getInstanceInfo(ContractAddress.create(BigInt(contractIndex)));
```

packages/browser-wallet/src/popup/store/token.ts
```diff
@@ -136,14 +137,22 @@ const cbf = atomFamily<string, Atom<ContractBalances>>((identifier: string) => {
...
+
+                let balances = {};
+                if (instanceInfo !== undefined) {
+                    balances = await getContractBalances(
+                        client,
+                        {
+                            index: BigInt(contractIndex),
+                            subindex: 0n,
+                            contractName: instanceInfo.name.value.substring(5),
```
orhoj, line 149: We have a helper function `getContractName` for this.

packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -410,52 +405,42 @@ export async function getTokens(
...
-        contractDetails.index,
-        contractDetails.subindex,
-        ids,
-        account,
-        onError
+    const tokenData: (TokenData | undefined)[] = await Promise.all(
+        ids.map(async (id, index) => {
+            let metadataUrl;
+            try {
+                [metadataUrl] = await getTokenUrl(client, [id], contractDetails);
```
orhoj, line 412: I think `getTokenUrl` is only ever invoked with a singleton list, so I would change the function to only support that.

orhoj (changes requested):

packages/browser-wallet/src/popup/pages/Account/Tokens/ManageTokens/state.ts
```diff
@@ -65,7 +65,7 @@ export const contractTokensAtom = (() => {
                     set(loadingAtom, true);
 
                     const tokens = get(listAtom);
-                    const topId = [...tokens].reverse()[0]?.pageId;
+                    const topId = tokens[tokens.length - 1]?.pageId;
```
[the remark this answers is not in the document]
shjortConcordium: The reason is to not reverse the list just to take the last element :thinking:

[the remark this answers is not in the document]
orhoj: Sure. I wonder why it was like that in the first place :smile:

packages/browser-wallet/src/shared/utils/token-helpers.ts
shjortConcordium: I just removed it, and used the helper from the SDK (did the same for getting balance)

shjortConcordium:
> The balance shows incorrectly when adding a token (it shows as 0 even if I have tokens).

I'm unable to reproduce this? :thinking:

orhoj (approved):

PR #416, opened 2023-12-13
Use new transaction fromJSON and toJSON

lassemoldrup:
## Purpose

Implements the changes described in https://github.com/Concordium/concordium-node-sdk-js/issues/307.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/shared/utils/payload-helpers.ts
+++ b/packages/browser-wallet/src/shared/utils/payload-helpers.ts
@@ -12,6 +12,7 @@ import {
     EntrypointName,
     Parameter,
     jsonParse,
+    getAccountTransactionHandler,
 } from '@concordium/web-sdk';
 import { Buffer } from 'buffer/';
 import { SmartContractParameters, SchemaType, SchemaWithContext } from '@concordium/browser-wallet-api-helpers';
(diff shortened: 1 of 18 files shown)


packages/browser-wallet/src/shared/utils/payload-helpers.ts
```diff
@@ -116,6 +119,11 @@ export function parsePayload(
                 },
             };
         }
+        case AccountTransactionType.Transfer:
+            return {
+                type,
+                payload: payload as SimpleTransferPayload,
+            };
```
soerenbf, line 126: Why not just the default case?

soerenbf (commented):
We should wait with merging this until the web-sdk version containing the functionality required is officially released.

Otherwise looks good! 🙂

packages/browser-wallet/src/shared/utils/payload-helpers.ts
lassemoldrup: The type checker couldn't figure out that it works without this, because `Transfer` is a special case in the `HeadlessTransaction` type I guess.

soerenbf (approved):
Looks good!

shjortConcordium (approved):

PR #419, opened 2024-01-11
Example of bigint in smart contract parameter failing.

limemloh:
## Purpose

_This should not be merged._

Changing the wCCD example to use a `bigint` as part of a smart contract parameter, currently causing the browser wallet to throw.

It seems the wallet does not support using `bigint` as part of a parameter for a smart contract, meaning it is impossible to use integers that are larger than `number`. This example is showcasing it using a contract address which uses `u64`, but the issue also applies for contracts just using u64 as part of their parameter.


limemloh:
Created an issue #420 tracking this

shjortConcordium:
Thank you, this was fixed (by yourself)

PR #421, opened 2024-01-22
Fix: Error when using bigint in smart contract parameters

### LINKED ISSUE #420: Error when using `bigint` in smart contract parameters
**Bug Description**
The wallet does not support using `bigint` as part of a parameter for a smart contract, meaning it is impossible to use integers that are larger than supported by `number`. This example is showcasing it using a contract address which uses `u64`, but the issue also applies for contracts just using `u64` as part of their parameter.

**Steps to Reproduce**
Use a `bigint` for any parameter which can be bigger than `number`, such as `u64` or index in a smart contract address.
An example can be found in the branch [`bigint-in-sc-params`](https://github.com/Concordium/concordium-browser-wallet/tree/bigint-in-sc-params)

**Expected Result**
Success

**Actual Result**
Error produced by the wallet API

limemloh:
## Purpose

Closes #420.

## Changes

- The injected script will now send the smart contract parameters _stringified_ using `json-bigint`, which is then parsed in the extension when parsing the payload. This is all internal and should not result in any breaking changes.
- The type `SmartContractParameters` is extended with `bigint`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api-helpers/CHANGELOG.md
+++ b/packages/browser-wallet-api-helpers/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Fixed
+
+-   Extend type `SmartContractParameters` with `bigint` allowing large numbers in smart contract parameters.
+
 ## 3.0.0
 
 ### Breaking changes
--- a/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
+++ b/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
@@ -70,6 +70,7 @@ export type SmartContractParameters =
     | { [key: string]: SmartContractParameters }
     | SmartContractParameters[]
     | number
+    | bigint
     | string
     | boolean;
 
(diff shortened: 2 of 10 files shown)


shjortConcordium (commented):
I think I would prefer using our internal stringify/parse instead of adding a dependency.
It breaks the display in the popup.
otherwise LGTM.

Screenshot from 2024-01-22 16-06-00
Screenshot from 2024-01-22 16-03-53-1

limemloh:
> I think I would prefer using our internal stringify/parse instead of adding a dependency. 

The issue with the internal stringify/parse is that it does not support arbitrary JSON, ex. if my smart contract takes the JSON `{ '@type': "bigint", value: "5"}` then it would hit our internal representation and be replaced with `5`.

limemloh:
> It breaks the display in the popup.
I have fixed that part again

shjortConcordium (approved):

PR #429, opened 2024-01-30
Wrong identity providers

orhoj:
## Purpose
Fix a bug where recovery from the wallet settings menu would use a list of identity providers from the wrong network.

## Changes
- Reset the identity provider atom when updating the network.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/package.json
+++ b/packages/browser-wallet/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@concordium/browser-wallet",
     "private": true,
-    "version": "1.3.2",
+    "version": "1.4.0",
     "description": "Browser extension wallet for the Concordium blockchain",
     "author": "Concordium Software",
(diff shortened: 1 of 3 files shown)


packages/browser-wallet/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@concordium/browser-wallet",
     "private": true,
-    "version": "1.3.2",
+    "version": "1.4.0",
```
soerenbf, line 4: Would this not be a patch release?

soerenbf (approved):

packages/browser-wallet/package.json
orhoj: Yes, but what you don't see is that I have already merged in some new functionality in a different PR. I just bumped the version here to avoid another PR.

PR #432, opened 2024-02-05
Fix ChooseIdentity component using wrong identity index

orhoj:
## Purpose
Fix an issue where the wrong identity was used for creating an account if the wallet contained failed identities.

## Changes
- Gather all identities to be able to set the correct identity storage index, so that on the following page the correct identity is shown as the selected one.

Note that this is just a quickfix. In general the fact that this is controlled by a global value in storage is bad, and ideally we would refactor it.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/pages/AddAccount/ChooseIdentity.tsx
+++ b/packages/browser-wallet/src/popup/pages/AddAccount/ChooseIdentity.tsx
@@ -5,7 +5,7 @@ import { useAtomValue, useSetAtom } from 'jotai';
 import IdentityProviderIcon from '@popup/shared/IdentityProviderIcon';
 
 import IdCard from '@popup/shared/IdCard';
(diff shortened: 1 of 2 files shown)


packages/browser-wallet/src/popup/pages/AddAccount/ChooseIdentity.tsx
```diff
@@ -64,6 +64,11 @@ export default function ChooseIdentity() {
             </Modal>
             <p className="m-t-20 text-center p-h-10">{t('chooseIdentity')}</p>
             {identities.value.map((identity, i) => {
+                // Do not allow the user to select an un-confirmed identity.
+                if (!isConfirmedIdentity(identity)) {
+                    return null;
+                }
```
shjortConcordium, line 70: What a nice solution! :)

shjortConcordium (approved):

PR #440, opened 2024-03-07
Fullscreen view

orhoj:
## Purpose
Add a button that opens the browser wallet in fullscreen mode (in a tab).

## Changes
- Added button in settings that will open the wallet in a tab instead of the extension popup window.
- Adjusted the UI of the wallet to work better in the fullscreen mode (centered, some margins, hide some overflows of menus, hide the connected box, go to identities page on issuance end, fix issue with web3id import popup window).

## Checklist

      hard-to-understand areas.


shjortConcordium (approved):
Could be nice to change to the tab, or make a notification/twinkle, when we return from identityIssuance/get requests from dapps. (But that is probably gold plating here)

PR #446, opened 2024-03-14
Check that docker files can be built in CI

shjortConcordium:
## Purpose

Ensure the docker files are not "silently broken". :upside_down_face:

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -61,6 +61,24 @@ jobs:
             - name: TypeCheck
               run: yarn type:check
 
+    build-example-docker:
+        runs-on: ubuntu-20.04
+        strategy:
+            matrix:
+                image:
+                    - { dir: eSealing, dockerfile: Dockerfile }
+                    - { dir: voting, dockerfile: Dockerfile }
+                    - { dir: wCCD, dockerfile: Dockerfile, extraArguments: '--build-arg NETWORK="mainnet"' }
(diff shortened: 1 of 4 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -61,6 +61,24 @@ jobs:
...
               run: yarn type:check
 
+    build-example-docker:
+        runs-on: ubuntu-20.04
+        strategy:
+            matrix:
+                image:
+                    - { dir: eSealing, dockerfile: Dockerfile }
+                    - { dir: voting, dockerfile: Dockerfile }
+                    - { dir: wCCD, dockerfile: Dockerfile, extraArguments: '--build-arg NETWORK="mainnet"' }
```
DOBEN, line 71: Since we host both `mainnet/testnet` version of the dapp. I suggest we just test building the image for both environmental variables in the CI pipeline as well.

```suggestion
                    - { dir: wCCD, dockerfile: Dockerfile, extraArguments: '--build-arg NETWORK="mainnet"' }
                    - { dir: wCCD, dockerfile: Dockerfile, extraArguments: '--build-arg NETWORK="testnet"' }
```

DOBEN (approved):

.github/workflows/build-lint-test.yml
abizjak: To me it seems excessive to test the build with both versions of the value.

PR #450, opened 2024-03-22
(Partially) Remove 30000

shjortConcordium:
## Purpose

Remove some of the hardcoded 30000 energy on contract updates.

## Changes

Removed 30000 energy from updates in the wCCD, voting and eSealing dApps

## Checklist

      hard-to-understand areas.


examples/eSealing/src/utils.ts
```diff
@@ -8,23 +18,46 @@ import { E_SEALING_CONTRACT_NAME, E_SEALING_RAW_SCHEMA } from './constants';
...
         account,
         AccountTransactionType.Update,
         {
             amount: CcdAmount.fromMicroCcd(0),
-            address: ContractAddress.create(index, subindex),
-            receiveName: ReceiveName.fromString(`${E_SEALING_CONTRACT_NAME}.registerFile`),
-            maxContractExecutionEnergy: Energy.create(30000n),
+            address: contractAddress,
+            receiveName,
+            maxContractExecutionEnergy: invokeResult.usedEnergy,
```
DOBEN, line 56: Adding a small margin `EPSILON_ENERGY` will help in cases where someone else uses the front end at the same time and the state of the smart contract has changed between the simulation and the execution of the transaction.
```
    const maxContractExecutionEnergy = Energy.create(dryRunResult.usedEnergy.value + EPSILON_ENERGY);

```

https://github.com/Concordium/concordium-dapp-examples/blob/main/trackAndTrace/frontend/src/track_and_trace_contract.tsx#L63

Applies to the other examples below as well.

examples/eSealing/src/utils.ts
```diff
@@ -8,23 +18,46 @@ import { E_SEALING_CONTRACT_NAME, E_SEALING_RAW_SCHEMA } from './constants';
...
+
+    const invokeResult = await grpcClient.invokeContract({
+        contract: contractAddress,
+        method: receiveName,
+        invoker: AccountAddress.fromBase58(account),
+        parameter: serializedParameters,
+    });
+
+    if (invokeResult.tag === 'failure') {
+        throw Error('Transaction would fail!');
```
DOBEN, line 46: It would be helpful if that error message reports the reason for the failure. The following, I suggest to cover: 

- Return the `rejectReason` of the smart contract if present.
- Return the `invokeResult.tag` (if it is not `failure`) since that is where we can see `OutOfEnergy` errors:
   https://github.com/Concordium/concordium-smart-contract-tools/blob/main/front-end-tools/src/utils.ts#L209


```
    const fullEntryPointName = `${contractName.value}.${entryPoint.value}`;

    if (!res || res.tag === 'failure') {
        const rejectReason = JSON.stringify(
            ((res as InvokeContractFailedResult)?.reason as RejectedReceive)?.rejectReason
        );

        throw new Error(
            `RPC call 'invokeContract' on method '${fullEntryPointName}' of contract '${contractIndex}' failed.
            ${rejectReason !== undefined ? `Reject reason: ${rejectReason}` : ''}`
        );
    }
```

Reason for adding this reporting of errors now:
Before, people who reported problems while using this front end were able to give us a "failedTransactionHash" that we could use for further investigation. We don't have this anymore, so reporting potential problems here is what we have to rely on now.

This comment applies to the other examples as well.

Ivan-Mahda:
Code applied during examples migration to dApp-examples
In PRs:
https://github.com/Concordium/concordium-dapp-examples/pull/149
https://github.com/Concordium/concordium-dapp-examples/pull/151

PR #459, opened 2024-04-02
[BRO-31] Errors on websites with strict Content Security Policy

Ivan-Mahda:
## Purpose

The injected script attempts to compile and instantiate WebAssembly, which fails when a website uses a Content Security Policy which does not allow for running WebAssembly.
Main reason esbuild does not provide proper tree-shaking and as result unused WebAssembly functions included in to bundle
This can be avoided by importing only functions you need, without importing all module.

## Changes

Created yarn patch for @concordium/web-sdk
Now it is possible to import from ./types/ directory, and avoid tree-shaking bug of esbuild

## Checklist

      hard-to-understand areas.


abizjak:
@Ivan-Mahda I will take myself off the reviewers list. Soren and Emil are the experts anyhow.

soerenbf (approved):
Great detective work @Ivan-Mahda!

It seems that this completely removes the WASM error.. I actually thought we needed to dive into the legacy grpc client and well and do some updates, but seems like that is not needed anyway 🤷

.yarn/patches/@concordium-web-sdk-npm-7.3.2-42c2804ca6.patch
```diff
@@ -0,0 +1,130 @@
...
+index 2653ea6fbf48c58e34e4b8235fc624d2a358b2e0..e4cbc990fa038fb69504ef7cb28eb83c82178c2b 100644
+--- a/package.json
++++ b/package.json
+@@ -38,6 +38,11 @@
+             "react-native": null,
+             "default": "./lib/esm/pub/wasm.js"
+         },
++        "./types/*": {
++            "types": "./lib/esm/types/*.d.ts",
++            "bun": "./src/pub/*.ts",
```
limemloh, line 125: **Suggestion:** I don't think "bun" is right.
It doesn't really matter for this, since we don't use it here, but we probably want this in the JS SDK, where it needs to be correct.

```suggestion
+            "bun": "./src/types/*.ts",
```

.yarn/patches/@concordium-web-sdk-npm-7.3.2-42c2804ca6.patch
```diff
@@ -0,0 +1,130 @@
...
++      <updated>1712067131693</updated>
++      <workItem from="1712067132751" duration="29000" />
++    </task>
++    <servers />
++  </component>
++  <component name="TypeScriptGeneratedFilesManager">
++    <option name="version" value="3" />
++  </component>
++</project>
+\ No newline at end of file
```
limemloh, line 114: **Request change:** The changes above this point seems irrelevant to include in the patch, could you clean this up a bit?

limemloh (commented):
The patch is nice for now, but I think we should fix it in the SDK straight away as well and then remove this patch again.

soerenbf:
> The patch is nice for now, but I think we should fix it in the SDK straight away as well and then remove this patch again.

I suggested we do this as a `yarn patch` initially, as I thought we might want to discuss if we want to make the types accessible in this manner in the SDK.

One point of view could be that it worsens the user experience, as `fromJSON` would trigger n (i.e. the number of wrapper types) suggestions from the typescript LSP along with the code produced being less readable (but maybe these rules are not for us to impose upon others?).

Of course another point of view could be that it enables people to use tools like ESBuild and potentially other tools with less ability to tree-shake than the state-of-the-art bundlers.

Personally, I feel inclined towards the former of the two.. But I don't have any strong opinion on this matter tbh.

limemloh:
> > The patch is nice for now, but I think we should fix it in the SDK straight away as well and then remove this patch again.
> 
> I suggested we do this as a `yarn patch` initially, as I thought we might want to discuss if we want to make the types accessible in this manner in the SDK.
> 
> One point of view could be that it worsens the user experience, as `fromJSON` would trigger n (i.e. the number of wrapper types) suggestions from the typescript LSP along with the code produced being less readable (but maybe these rules are not for us to impose upon others?).
> 
> Of course another point of view could be that it enables people to use tools like ESBuild and potentially other tools with less ability to tree-shake than the state-of-the-art bundlers.
> 
> Personally, I feel inclined towards the former of the two.. But I don't have any strong opinion on this matter tbh.
> 
> What do you think @Ivan-Mahda and @limemloh ?

Just wanted to check the experience if we did this, and you are right, it does add a lot of noise:
image
I agree then, let's get this merged and discuss find the best solution for the SDK after

Issue #279, opened 2023-04-18
Payload decoding in the browser wallet of the CIS3 standard


DOBEN:
**Task description**
If a smart contract implements the CIS3 standard (sponsored transactions). The sponsored transaction `message` will be sent to the Concordium browser wallet to be signed by the user. Currently, when providing the `schema` to the browser wallet, it can decode the object except for the payload field which is an array of bytes.

Feature to be added to the browser wallet:
The browser wallet can also decode the payload field by looking up the parameter schema of the `entry_point_name` and using this schema to decode the payload. The payload is the input parameter that is sent to the `entry_point_name`. 

![DecodingOfSigningOfMessage](https://user-images.githubusercontent.com/13039169/232790023-a950347b-ef65-42f4-b4ab-584641161c6b.png)

CIS3 standard:
https://github.com/Concordium/concordium-update-proposals/pull/41

limemloh:
Since the current approach is to use `signMessage` function provided by the wallet, we don't really know what we are signing, and we can't really deserialize the message and look whether it is for CIS-3 reliably.
A solution could be to extend the wallet API with `signCis3Message` which is tailored for this.

abizjak:
https://github.com/Concordium/concordium-dapp-examples/issues/77

PR #460, opened 2024-04-04
[BRO-13] Payload decoding in the browser wallet of the CIS3 standard

Ivan-Mahda:
## Purpose

The browser wallet can also decode the payload field by looking up the parameter schema of the entry_point_name and using this schema to decode the payload. The payload is the input parameter that is sent to the entry_point_name.

## Changes

Proposal on how to decode the payload for the CIS3 contract.
Screenshot from 2024-04-15 15-32-01

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
+++ b/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
@@ -258,7 +258,7 @@ interface MainWalletApi {
      * @param contractAddress the {@link ContractAddress} of the contract
      * @param contractName the {@link ContractName} of the contract
      * @param entrypointName the {@link EntrypointName} of the contract
-     * @param nonce the revocation nonce
+     * @param nonce the nonce (CIS3 standard) that was part of the message that was signed
      * @param expiryTimeSignature RFC 3339 format (e.g. 2030-08-08T05:15:00Z)
      * @param accountAddress the address of the account that should sign the message
      * @param payloadMessage payload message to be signed, complete CIS3 message will be created from provided parameters. Note that the wallet will prepend some bytes to ensure the message cannot be a transaction. The message should be { @link SignMessageObject }.
@@ -269,7 +269,7 @@ interface MainWalletApi {
         contractAddress: ContractAddress.Type,
         contractName: ContractName.Type,
         entrypointName: EntrypointName.Type,
-        nonce: bigint,
+        nonce: bigint | number,
         expiryTimeSignature: string,
         accountAddress: AccountAddressSource,
         payloadMessage: SignMessageObject
--- a/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
+++ b/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
@@ -254,12 +254,14 @@ interface MainWalletApi {
 
     /**
      * Sends a message of the CIS3 contract standard, to the Concordium Wallet and awaits the users action. If the user signs the message, this will resolve to the signature.
-     * Note that if the user rejects signing the message, this will throw an error.
+     * 
      * @param contractAddress the {@link ContractAddress} of the contract
      * @param contractName the {@link ContractName} of the contract
      * @param entrypointName the {@link EntrypointName} of the contract
      * @param accountAddress the address of the account that should sign the message
      * @param message message to be signed. Note that the wallet will prepend some bytes to ensure the message cannot be a transaction. The message should be { @link SignMessageObject }.
+     *
+     * @throws if the user rejects signing the message.
      */
     signCIS3Message(
         contractAddress: ContractAddress.Type,
--- a/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
+++ b/packages/browser-wallet-api-helpers/src/wallet-api-types.ts
@@ -266,7 +266,7 @@ interface MainWalletApi {
         contractName: ContractName.Type,
         entrypointName: EntrypointName.Type,
         accountAddress: AccountAddressSource,
-        message: string | SignMessageObject
+        message: SignMessageObject
     ): Promise<AccountTransactionSignature>;
 
     /**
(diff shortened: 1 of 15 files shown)


limemloh:
As mentioned in the related issue https://github.com/Concordium/concordium-browser-wallet/issues/279 
We cannot make `signMessage` behave like this, since it won't work reliably and worse it could be triggered for unrelated messages.

limemloh:
Adding @soerenbf as reviewer since he is more familiar with the UI part of the wallet than me

packages/browser-wallet/src/popup/pages/SignCIS3Message/SignCIS3Message.tsx
```diff
@@ -0,0 +1,176 @@
...
+            cis3ContractDetails: Cis3ContractDetailsObject;
+        };
+    };
+}
+
+type Cis3ContractDetailsObject = {
+    contractAddress: ContractAddress.Type;
+    contractName: ContractName.Type;
+    entrypointName: EntrypointName.Type;
+};
```
limemloh, line 52: **Suggestion:** Display the account address as well.

I think it would be good to be more explicit on which account address is used for signing.

packages/browser-wallet-api-helpers/src/wallet-api-types.ts
```diff
@@ -250,6 +252,23 @@ interface MainWalletApi {
...
+     * @param entrypointName the {@link EntrypointName} of the contract
+     * @param accountAddress the address of the account that should sign the message
+     * @param message message to be signed. Note that the wallet will prepend some bytes to ensure the message cannot be a transaction. The message should be { @link SignMessageObject }.
+     */
+    signCIS3Message(
+        contractAddress: ContractAddress.Type,
+        contractName: ContractName.Type,
+        entrypointName: EntrypointName.Type,
+        accountAddress: AccountAddressSource,
+        message: SignMessageObject
```
limemloh, line 269: **Request change:** Add missing parameter for `nonce` and `expiry` all part of the `PermitMessage` for the parameter of the [`permit` function in CIS-3](https://proposals.concordium.software/CIS/cis-3.html#parameter)

packages/browser-wallet/src/popup/pages/SignCIS3Message/SignCIS3Message.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+async function parseMessage(
+    message: SignMessageObject,
+    client: ConcordiumGRPCClient,
+    cis3ContractDetails: Cis3ContractDetailsObject
+) {
+    const deserializedMessage = deserializeTypeValue(
+        Buffer.from(message.data, 'hex'),
+        Buffer.from(message.schema, 'base64')
```
limemloh, line 65: **Request change:** For `signCIS3Message` the schema provided as part of `message` by the dapp should be the payload and payload schema only.

Current implementation is assuming the user provide the schema for the `PermitMessage` in CIS-3, but the `message` should only represent the payload part of it, the rest of the information is provided as arguments.

packages/browser-wallet/src/popup/pages/SignCIS3Message/SignCIS3Message.tsx
```diff
@@ -0,0 +1,176 @@
...
+    ) as DeserializedMessageObject;
+
+    const { contractAddress, contractName, entrypointName } = cis3ContractDetails;
+    const instanceInfo = await client.getInstanceInfo(contractAddress);
+
+    const schema = await client.getEmbeddedSchema(ModuleReference.fromHexString(instanceInfo.sourceModule.moduleRef));
```
limemloh, line 71: **Suggestion:** Ignore the embedded schema for now and assume the dapp always provide it (part of `message`), since this is what is done everywhere else in the wallet.

limemloh (changes requested):

soerenbf (commented):
I don't know that much about the CIS3 standard; I think @limemloh is better suited to comment on this part (as he already has). It would however be good if you created a small example application under `/examples` that uses the functionality.

There is an app `/examples/two-step-transfer` which has a pile of different functionality all inlined in a `<script>` tag in the html document. You can also find the sponsored transaction example there and adapt it. I would prefer if you created a small isolated (typescript) example though. We usually use `vite` to do this.

packages/browser-wallet/src/popup/pages/SignCIS3Message/SignCIS3Message.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const { contractAddress, contractName, entrypointName, nonce, expiryTimeSignature } = cis3ContractDetails;
+    const [parsedMessage, setParsedMessage] = useState<string>('');
+    const expiry = new Date(expiryTimeSignature).toString();
+
+    useEffect(() => {
+        try {
+            parseMessage(payloadMessage).then((m) => setParsedMessage(m));
+        } catch (e) {
+            setParsedMessage(t('unableToDeserialize'));
+        }
```
limemloh, line 97: **Suggestion:** Use `.catch` here.

To my understanding this error handling is unreachable even on error, since `parseMessage` is defined using `async` it will reject the promise instead, and since we don't use `await` in this function, this rejection is never turned into an exception.
I think the simplest solution is to use `.catch` instead of `try-catch` here.
```suggestion
        parseMessage(payloadMessage)
          .then((m) => setParsedMessage(m))
          .catch((_) => setParsedMessage(t('unableToDeserialize')));
```

limemloh (approved):
Looks good to me

soerenbf:
Are we not doing the example for running this, or what is the resolution on that?

Ivan-Mahda:
> Are we not doing the example for running this, or what is the resolution on that?

Yes, after wallet update. I will update example in sponsored transaction https://github.com/Concordium/concordium-dapp-examples/issues/77

PR #464, opened 2024-05-02
[BRO-29] Better error reporting in the browser wallet in case TokenMetadata

Ivan-Mahda:
## Purpose
Changed error message. Now instead of not showing invalid tokens, they displayed in token list with corresponding error. In order to show user, that we found tokens in contract, but they have error. Instead, just not showing them.

image

## Changes

Error message displayed in token list. Previously it was shown only in console.log

## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -33,6 +33,7 @@ export interface ContractDetails {
 
 export type ContractTokenDetails = TokenIdAndMetadata & {
     balance: bigint;
+    error: string;
```
limemloh, line 36: It would be good to document that the empty string is used for no error

limemloh (commented):
Please update the changelog as well

packages/browser-wallet/src/shared/utils/token-helpers.ts
soerenbf: I think it would be better to make it optional, as that explains the intention. Unless there's a reason for the empty string I'm missing?

packages/browser-wallet/src/popup/shared/ContractTokenLine/ContractTokenLine.tsx
```diff
@@ -38,15 +38,16 @@ export default function ContractTokenLine({
             className={clsx('contract-token-line__token', className)}
             style={style}
             onClick={() => onClick(token)}
+            disabled={!!token.error}
         >
             <div className="flex align-center h-full m-r-5">
                 <Img
-                    src={token.metadata.thumbnail?.url ?? token.metadata.display?.url ?? ''}
-                    alt={token.metadata.name ?? ''}
+                    src={token.metadata?.thumbnail?.url ?? token.metadata?.display?.url ?? ''}
```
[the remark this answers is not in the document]
Ivan-Mahda: Fixed `token.metadata` so it's never undefined

packages/browser-wallet/src/shared/utils/token-helpers.ts
Ivan-Mahda: `error` now optional

packages/browser-wallet/src/popup/pages/Account/Tokens/ManageTokens/ManageTokens.tsx
```diff
@@ -69,6 +69,7 @@ export default function AddTokens() {
                 accountTokens.value[contractDetails.index.toString()]?.map((token) => ({
                     ...token,
                     balance: currentContractBalances[token.id] ?? 0n,
+                    error: '',
```
soerenbf, line 72: Not needed with optional field

```suggestion
```

packages/browser-wallet/src/shared/utils/token-helpers.ts
```diff
@@ -347,32 +349,42 @@ export async function getTokens(
     );
     const tokenData: (TokenData | undefined)[] = await Promise.all(
         ids.map(async (id, index) => {
+            const internalData: TokenData = { id, metadataLink: '', metadata: {}, balance: 0n, error: '' };
```
soerenbf, line 352: And here

```suggestion
            const internalData: TokenData = { id, metadataLink: '', metadata: {}, balance: 0n };
```

soerenbf (approved):
As @limemloh suggested, we need to reflect the changes introduced with this in the changelog. Otherwise looks good to me 🙂

limemloh (approved):

PR #469, opened 2024-05-27
Remove check of redirectUri

limemloh:
## Purpose

When launching identity issuance, the browser wallet will send a request the ID provider and be redirected to the onboarding flow.
The browser wallet checks the redirected URL does not contain the `redirectUrl` which is later used for signalling the end of the flow, but there is no error handling for this case, so the wallet fails silently and is stuck on the page for generating the request.

One fix could be to produce an actual error, but this check seems arbitrary and provide no value, so instead I have removed it.
Also, it was hit by accident by an upcoming identity provider.
I have checked that this still works with every identity provider on testnet.

## Changes

- Remove the check of the redirected identity issuance url.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/CHANGELOG.md
+++ b/packages/browser-wallet/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+-   Remove check for redirectUri when launching identity issuance. This check was causing issues with an upcoming identity provider and seems to provide no value.
+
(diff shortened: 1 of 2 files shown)


Ivan-Mahda (approved):

soerenbf (approved):
I can't remember why this extra measure was in place, but everything works as expected for me as well when I test this.

PR #482, opened 2024-07-19
Fix voting example

Victor-N-Suadicani:
## Purpose

I'm not entirely certain what has happened, but it seems like some breaking changes were made to the web SDK and this example was never updated to still be functioning (when going to the result page, no results show up). Through some trial and error I managed to make it work again locally, although I'm not sure whether this is the "idiomatic" approach. Anyways, it seems to work.

## Changes

- The manifest and favicon are moved to the `public` folder instead of `public/static/img` - I'm not sure why they were in that nested folder, especially the manifest seeing as that is not even an image. The frontend seems to expect these files to be in the root? I'm not sure what's going on there.
- Changed various raw strings to use the expected typed parameter instead, as a breaking change to the web SDK seems to have made this a necessity.
- Changed the way the gRPC client is constructed, as the previous method is deprecated (this was indicated in a changelog entry somewhere).

## Checklist

      hard-to-understand areas.


DOBEN (commented):
Looks good.

Let's see if the plugin gets fixed soonish and we can remove the temporary fix with the `resolutions`:
https://github.com/Menci/vite-plugin-top-level-await/issues/52

DOBEN (approved):

PR #489, opened 2024-08-12
Fix inconsistent SDK version across processes.

soerenbf:
Also fixes compatibility layer not providing the assurance that the expected types are used (i.e. from the expected SDK version). This problem happened when newtypes created with another SDK version were fed to the wallet API but not recreated within the wallet API.

## Changes

- Wallet API now defines web-SDK as a peer dependency
- Ensures newtypes are re-created internally in the wallet API to ensure they match the expected version

## Checklist

      hard-to-understand areas.


packages/browser-wallet-api/package.json
```diff
@@ -21,6 +20,9 @@
         "jest": "^29.7.0",
         "ts-jest": "^29.1.1"
     },
+    "peerDependencies": {
+        "@concordium/web-sdk": "*"
+    },
```
soerenbf, line 25: I'm actually not 100% certain that this is enough to ensure consistency between the versions used in wallet-api/wallet..

I did 
```
yarn dedupe @concordium/web-sdk
```

which aligned the version across all packages.. As the browser wallet depends on this in `inject/index.ts` I would hope the browser wallet version is used, but I might be wrong...

Ivan-Mahda: Technically, it should work. 

But in the future, this may create additional complications with versions, somewhere there will be a `workspace^` and somewhere a version that can be checked only through `package.lock`.
In addition, there will be a risk that one version is installed locally and used during development, but during the build of the release package, updated one will be downloaded. Which adds complexity to package management.

soerenbf: Okay.. Do you have a better suggestion, or should we just go with this? One alternative approach I also had in mind was to refactor the workspace into

```
examples/
  ...
packages/
  browser-wallet/
    src/
      ...
      message-hub
      wallet-api
  browser-wallet-api-helpers/
    ...
```

This eliminates the need to define packages individually for wallet and wallet-api, and there is no real need for them to be defined as separate packages.

The reason I went with the current approach for now was simplicity, but I think the one outlined above would be safer

Ivan-Mahda (approved):

packages/browser-wallet-api/package.json
soerenbf: For context, structure proposed above was merged in with #490

PR #490, opened 2024-08-13
Refactor packages

soerenbf:


packages/browser-wallet/src/wallet-api/compatibility.ts
```diff
@@ -34,7 +34,7 @@ import {
     getAccountTransactionHandler,
     AccountTransactionPayloadJSON,
 } from '@concordium/web-sdk/types';
-import { empty } from '@concordium/web-sdk/types/Parameter';
+import { Parameter } from '@concordium/web-sdk';
```
limemloh, line 37: I think this change will include the wasm module again, since esbuild tree-shaking is not smart enough

limemloh (approved):
Great with the cleanup :partying_face: 
Only one thing we probably need to change

Ivan-Mahda (approved):

packages/browser-wallet/src/wallet-api/compatibility.ts
soerenbf: ah fuck.. good catch! I'll see if I can change it back.. changed it due to build errors

PR #543, opened 2024-10-22
UI update/amount picker

soerenbf:
## Purpose

Implementation of amount picker to be used for transfer flows + validation/delegation flows.

## Changes

- Fix storybook viewport sizes to the scales used in the wallet (+ fullscreen)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/browser-wallet/src/popup/popupX/shared/Form/TokenAmount/TokenAmount.tsx
+++ b/packages/browser-wallet/src/popup/popupX/shared/Form/TokenAmount/TokenAmount.tsx
@@ -1,0 +1,60 @@
+import React, { useMemo } from 'react';
+import { ContractAddress, HexString } from '@concordium/web-sdk';
+import { TokenMetadata } from '@shared/storage/types';
+import SideArrow from '@assets/svgX/side-arrow.svg';
+import ConcordiumLogo from '@assets/svgX/concordium-logo.svg';
+
+type Token =
+    | { type: 'ccd' }
+    | { type: 'cis2'; data: TokenMetadata; address: { id: HexString; contract: ContractAddress.Type } };
+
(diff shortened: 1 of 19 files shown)


packages/browser-wallet/src/popup/popupX/shared/Form/TokenAmount/TokenAmount.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { useState } from 'react';
+import { atomFamily, selectAtom, useAtomValue } from 'jotai/utils';
+import { AccountAddress, AccountInfo, ContractAddress, CIS2 } from '@concordium/web-sdk';
+import { atom } from 'jotai';
+
+import { useSelectedAccountInfo } from '@popup/shared/AccountInfoListenerContext/AccountInfoListenerContext';
+import { contractBalancesFamily } from '@popup/store/token';
+import { ensureDefined } from '@shared/utils/basic-helpers';
```
soerenbf, line 8: I can't figure out what the strategy is for all the things in `@popup/shared`. Do we move everything we use to `@popup/popupX/shared`? I reckon it would be much simpler to just keep everything in `@popup/shared` in its current place and run some tool to identify dead code when we're all done.

@Ivan-Mahda @limemloh what do you think?

limemloh: Moving things over could also be a source of conflicts, so my guess is also that it would be simpler to expand the current one

soerenbf: Good point

Ivan-Mahda (approved):

PR #558, opened 2024-10-29
UI update/update delegation

soerenbf:
## Checklist

      hard-to-understand areas.


packages/browser-wallet/src/popup/popupX/pages/EarningRewards/Delegator/TransactionFlow.tsx
```diff
@@ -0,0 +1,128 @@
...
+                return;
+            }
+
+            nav(pathname, { replace: true, state: form }); // Override current router entry with stateful version
+
+            const submitDelegatorState: DelegationResultLocationState = {
+                payload,
+                type: existingValues !== undefined ? 'change' : 'register',
+            };
+            nav(absoluteRoutes.settings.earn.delegator.submit.path, { state: submitDelegatorState });
```
limemloh, line 62: **Question:** I get an error when submitting is this expected?

The error is related to signing it seems `Uint8Array expected`

packages/browser-wallet/src/popup/popupX/pages/EarningRewards/Delegator/TransactionFlow.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { configureDelegatorPayloadFromForm, type DelegatorForm } from './util';
+import { DelegationResultLocationState } from './Result/DelegationResult';
+
+function NoChangesNotice(props: FullscreenNoticeProps) {
+    const { t } = useTranslation('x', { keyPrefix: 'earn.delegator.update.noChangesNotice' });
+    return (
+        <FullscreenNotice {...props}>
+            <Page>
+                <Page.Top heading={t('title')} />
+                {t('description')}
```
limemloh, line 25: **Suggestion:** Wrap in some text component, the font seems to be different at least.

packages/browser-wallet/src/popup/popupX/pages/EarningRewards/Delegator/TransactionFlow.tsx
```diff
@@ -0,0 +1,128 @@
...
+import Button from '@popup/popupX/shared/Button';
+import DelegatorStake from './Stake';
+import DelegatorType from './Type';
+import { configureDelegatorPayloadFromForm, type DelegatorForm } from './util';
+import { DelegationResultLocationState } from './Result/DelegationResult';
+
+function NoChangesNotice(props: FullscreenNoticeProps) {
+    const { t } = useTranslation('x', { keyPrefix: 'earn.delegator.update.noChangesNotice' });
+    return (
+        <FullscreenNotice {...props}>
```
limemloh, line 22: **Question (non-blocking):** I find the flow hard to follow since some of it is tracked in the path and some in modals, why is this even a modal and not just another page part of the routing? :smile:

limemloh (approved):

packages/browser-wallet/src/popup/popupX/pages/EarningRewards/Delegator/TransactionFlow.tsx
soerenbf: The point of these "notices" is to have the user confirm some action or simply show some information to the user. Routing to another page tears down the state built for that route/page, which would require you to save the state somehow to allow the user to return to the previous page and rebuild the page from the saved state.

We can completely avoid this by treating this as what it actually is (a prompt for the user) instead of what it looks like (a subsequential page in the flow) 😄

packages/browser-wallet/src/popup/popupX/pages/EarningRewards/Delegator/TransactionFlow.tsx
soerenbf: You're right, good catch 👍

PR #561, opened 2024-10-30
Hook up account selector on main page

limemloh:
## Purpose

Connect account selector on main page with actual data.

### DIFF
--- a/packages/browser-wallet/src/popup/popupX/page-layouts/MainLayout/Header/Header.scss
+++ b/packages/browser-wallet/src/popup/popupX/page-layouts/MainLayout/Header/Header.scss
@@ -161,9 +161,7 @@
         &_search-form {
             display: flex;
 
-            .account-search__form {
-                width: 100%;
-
+            .form-search {
                 .form-input__field {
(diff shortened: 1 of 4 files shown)


soerenbf (commented):
Screenshot for context:

Screenshot 2024-10-30 at 11 10 08

A couple of things:

- Explicitly named accounts looks... unfortunate, when selected.
  - CCD balance alignment does not match other rows
  - Account name wraps
  - I would discuss with the designer to figure out what to do here, if it's not easily solvable
  -  **suggestion:** `text-align: left` the text + move selected icon left with `position: absolute`
- The rows don't appear clickable when hovering them. Would be nice with an indication that this is the case (at minimum `cursor: pointer`).
  - Very nice that I can use the keyboard to navigate to each row though! 🎉 
- The scroll bar offsets the entire UI. I would expect only the list to scroll, as I think there is enough space to show a reasonable number of accounts even when subtracting the header of the layout (i.e. fullscreen/connected + 9dots nav button).

Ivan-Mahda (approved):

PR #566, opened 2024-11-08
Add create account pages

limemloh:
## Purpose

Implement account creation pages, i.e. select identity and confirm.

Some details that I have left out and will address in a future PR are:

- Missing the transparent background for confirm page.
- Support pending accounts in the home page, currently it is just displaying "Loading".


soerenbf (approved):
Well, in that case it looks good 😄

soerenbf (commented):
If I attempt to create an account with an expired identity, I don't get any feedback that the account creation did not succeed. 

In the current wallet I get an error (an obscure bad UX RPC error message, but hey.. it's something?). I think it would make sense to filter out identities that cannot be used for account creation.

packages/browser-wallet/src/popup/popupX/pages/CreateAccount/CreateAccount.tsx
```diff
@@ -10,9 +10,31 @@ import Button from '@popup/popupX/shared/Button';
...
+ */
+function useValidIdentities(): ConfirmedIdentity[] {
+    const identities = useAtomValue(identitiesAtom);
+    return useMemo(() => {
+        const now = new Date();
+        return identities.flatMap((id) => {
+            if (id.status !== CreationStatus.Confirmed) {
+                return [];
+            }
+            const validToDate = new Date(id.idObject.value.attributeList.validTo);
```
soerenbf, line 27: This does not seem to work, as the the dates are stored as "YYYYMM" which is not parsed properly by the `Date` constructor. I thought there was a helper function to parse this in `wallet-common-helpers/utils/timeHelpers.ts`, but I cannot seem to find it..

limemloh: ahh right, I forgot this was the format.
I guess my test then just happened then... I'll fix this. Good catch

soerenbf (approved):

PR #568, opened 2024-11-08
Main/home page pending accounts.

limemloh:
## Purpose

Right after completing account creation, the account will have status of pending.
This PR extends the home page to show information for pending accounts.


soerenbf (approved):
I'm not a fan of using the account balance as a header for showing that the account is pending. I don't think this is something we should argue about though, but I would ask for design clarification here.

If you want to merge it now, go ahead. But I think we should make an issue to request a design for this as it's a part of the required flow for any new user to go through.

PR #569, opened 2024-11-11
Show rejected ID cards

limemloh:
## Purpose

Show cards for rejected identities.

## Changes

- Minor refactor of `IdCard` component


packages/browser-wallet/src/assets/svgX/rejected.svg
soerenbf, line 1: I'm sure this svg is already there with another colour. Can we just use that and apply `$color-red-attention` to it?

packages/browser-wallet/src/popup/popupX/pages/Onboarding/IdSubmitted.tsx
```diff
@@ -21,7 +20,16 @@ export default function IdSubmitted() {
             <Page.Top heading={t('yourId')} />
             <Page.Main>
                 <Text.Capture>{t('idSubmitInfo')}</Text.Capture>
-                <IdCard rowsIdInfo={rowsIdInfo} idProviderName="TODO" identityName="TODO" />
+                <IdCard title="Identity 1" subtitle="Submitted to NotaBene">
```
soerenbf, line 23: Is this (i.e. "Identity 1", "Submitted by NotaBene") accurate?

packages/browser-wallet/src/popup/popupX/pages/Restore/RestoreResult.tsx
```diff
@@ -30,7 +29,17 @@ export default function RestoreResult() {
             <Page.Top heading={t('result')} />
             <Page.Main>
                 <Text.Capture>{t('recoveredIds')}</Text.Capture>
-                <IdCard rowsIdInfo={rowsIdInfo} idProviderName="TODO" identityName="TODO" />
+
+                <IdCard title="Identity 1" subtitle="Verified by NotaBene">
```
soerenbf, line 33: And here

packages/browser-wallet/src/popup/popupX/shared/IdCard/RejectedIdCard.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default function RejectedIdCard({ identity }: RejectedIdentityProps) {
+    const { t } = useTranslation('x', { keyPrefix: 'sharedX' });
+    const providers = useAtomValue(identityProvidersAtom);
+    const provider = providers.find((p) => p.ipInfo.ipIdentity === identity.providerIndex);
+    const idProviderName = provider?.ipInfo.ipDescription.name ?? 'Unknown';
+    return (
+        <IdCard title={identity.name} subtitle={t('idCard.rejectedBy', { idProviderName })}>
+            <IdCard.Content className="rejected">
+                <Rejected />
```
soerenbf, line 23: I don't know about the icon. It clearly states that there is an error, but I don't think it looks consistent when comparing with verified identities. IMO, removing the icon is better design-wise, but arguably less clear in terms of what's going on.

I would ask for design clarification here as well.

Personally, I would go for something like 
Screenshot 2024-11-13 at 09 45 51

But I don't think that looks fantastic either 🤷 

Also maybe it's redundant to state the rejection so many times as well? I think maybe remove it from the black part of the id card and just simply state the error detail received from the IDP?

packages/browser-wallet/src/popup/popupX/pages/Onboarding/IdSubmitted.tsx
limemloh: These are again just placeholder text until we implement the onboarding flow.

packages/browser-wallet/src/popup/popupX/pages/Restore/RestoreResult.tsx
limemloh: lik above

packages/browser-wallet/src/popup/popupX/shared/IdCard/RejectedIdCard.tsx
limemloh: I agree the design could be improved, but I would suggest we iterate on it in the design document and stick to this for now.

soerenbf (approved):

